PR #30, opened 2024-09-17
:card_file_box: [#14] added informatie categorie model

### LINKED ISSUE #14: ODRC: implement informatiecategorie resource

```yaml
InformatieCategorie:
  type: object
  properties:
    identifier:
      type: string
      format: uri
    order:
      type: number  # integer
    name:
      type: string
    namePlural:
      type: string
    definition:
      type: string
    origin:
      type: string
      enum:
        - waardelijst
```

    - Order by `order`
    - List filter on `origin`
    - Paginate list endpoint

bart-maykin:
Partially Fixes #14

**Changes**

Aded informatie categorie model and migration


src/woo_publications/api/constants.py
```diff
@@ -0,0 +1,7 @@
+from django.db.models import TextChoices
+from django.utils.translation import gettext_lazy as _
+
+
+class InformatieCategorieChoices(TextChoices):
```
sergei-maertens, line 5:
```suggestion
class InformatieCategorieOrigins(TextChoices):
```

src/woo_publications/api/constants.py
```diff
@@ -0,0 +1,7 @@
...
+    waardelijst = "waardelijst", _("Waardelijst")
+    zelf_toegevoegd = "zelf_toegevoegd", _("Zelf toegevoegd")
```
sergei-maertens, line 7: I know it's a bit of an extra hurdle, but please define translatable string literals in English, and we'll manage the translations with Django's makemessages/compilemessages.

```suggestion
    waardelijst = "waardelijst", _("Value list")
    zelf_toegevoegd = "zelf_toegevoegd", _("Custom entry")
```

src/woo_publications/api/models/__init__.py
sergei-maertens, line 1: I think I'd rather avoid dumping all the models in the `api` "app", but split that out in two packages:

* `woo_publications.metadata` -> contains the model/admin etc. for `InformatieCategorie`, `Theme`...
* `woo_publications.publications` -> contains the `Publication` and `Document` models

That way, in the code we "physically" separate the publication content and the stuff that is supporting it (metadata only exists so that publications can be realized)

src/woo_publications/api/models/category.py
```diff
@@ -0,0 +1,32 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from ordered_model.models import OrderedModel
+
+from ..constants import InformatieCategorieChoices
+
+
+class InformatieCategorie(OrderedModel):
+    identifier = models.URLField(_("Identifier"), max_length=255, unique=True)
```
sergei-maertens, line 10: please make sure that all `verbose_name`'s start with a lower case letter - Django takes care of capitalizing that at the form layer

```suggestion
    identifier = models.URLField(_("identifier"), max_length=255, unique=True)
```

src/woo_publications/api/models/category.py
```diff
@@ -0,0 +1,32 @@
...
+    name = models.CharField(_("Name"), max_length=80)
+    name_plural = models.CharField(_("Name plural"), max_length=80)
```
sergei-maertens, line 12:
```suggestion
    name_plural = models.CharField(_("Name plural"), max_length=80, blank=True)
```
I thought you mentioned yesterday that this is not guaranteed to be provided

src/woo_publications/api/models/category.py
```diff
@@ -0,0 +1,32 @@
...
+    definition = models.TextField(_("Definition"), blank=True)
+    origin = models.CharField(
+        _("Origin"),
+        choices=InformatieCategorieChoices.choices,
+        blank=False,
+        max_length=15,
+    )
```
sergei-maertens, line 19:
```suggestion
    origin = models.CharField(
        _("Origin"),
        choices=InformatieCategorieChoices.choices,
        blank=False,
        max_length=15,
        default=InformatieCategorieOrigins.zelf_toegevoegd,
    )
```
as discussed yesterday?

Can you also see if we can think of a proper English term for `zelf_toegevoegd` in the code? The value like that is fine so that we have dutch strings in the API spec. Maybe we can go with `user_content` or `custom_entry`?

src/woo_publications/api/models/category.py
```diff
@@ -0,0 +1,32 @@
...
+
+    class Meta:
+        verbose_name = _("informatie categorie")
+        verbose_name_plural = _("informatie categorieën")
+        ordering = ("order",)
```
sergei-maertens, line 24:
```suggestion
    class Meta(OrderedModel.Meta):
        verbose_name = _("informatie categorie")
        verbose_name_plural = _("information categories")
```
* the `Meta` inheritance takes care of setting up the ordering properly, it's a shame they don't document this properly.
* Code base must be English - the only exception is stuff that directly ends up in the API spec since we want that interface to be in Dutch to avoid people bikeshedding about that. Now I'm not sure if we should define the model fields in English or not - I *think* at the API level you can use the serializer `source` option via `Serializer.Meta.extra_kwargs` but if that is not an obvious solution, let's just use Dutch field names

src/woo_publications/api/models/category.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def save(self, **kwargs):
+        if not self.origin:
+            self.origin = InformatieCategorieChoices.zelf_toegevoegd
```
sergei-maertens, line 28: that should be handled by the `default=` model option and the save override shouldn't be needed then.

Also, the model save doesn't return anything, so you can do `super().save(...)` instead of `return super().save(...)`

src/woo_publications/api/models/category.py
```diff
@@ -0,0 +1,32 @@
...
+        return super().save(**kwargs)
+
+    def __str__(self):
+        return self.identifier
```
sergei-maertens, line 32: Please return the human readable name instead of a system value

```suggestion
    def __str__(self):
        return self.name
```

sergei-maertens (changes requested):

src/woo_publications/api/models/category.py
bart-maykin: Is this really the best option? Because we aren't stopping anyone from using the same name twice. Should we add the identifier within brackets after the name when the name is duplicate or something like that?

sergei-maertens: Yes, it's just used for informative messages in the admin like "Succesfully changed the information categorie foobarbaz". The `__str__` method is intended for a quick, recognizable representation of the model instance :)

src/woo_publications/metadata/models.py
```diff
@@ -0,0 +1,48 @@
...
+    definitie = models.TextField(
+        _("definition"),
+        help_text=_("The description of the category."),
+        blank=True,
+    )
+    oorsprong = models.CharField(
+        _("origin"),
+        help_text=_(
+            "Whenever the category is defined by the government or is custom made."
+        ),
```
sergei-maertens, line 36:
```suggestion
        help_text=_(
            "Determines where the category is defined and sourced from, and how "
            "the identifier should be interpreted. If the value list is the origin, the "
            "category can not be modified or deleted."
        ),
```

sergei-maertens (approved):

PR #31, opened 2024-09-17
:sparkles: [#14] added admin page for informatie categorie

### LINKED ISSUE #14: ODRC: implement informatiecategorie resource

bart-maykin:
Partially Fixes #14

**Changes**

Added admin page for informatie categorie
Updated the admin default index


src/woo_publications/metadata/tests/test_informatie_categorie_admin.py
```diff
@@ -0,0 +1,212 @@
...
+        self.assertEqual(response.status_code, 200)
+        # page doesn't have save button.
+        with self.assertRaisesMessage(
+            AssertionError,
+            'Couldn\'t find \'<input class="default" name="_save" type="submit" value="Opslaan">\' in response',
+        ):
+            self.assertInHTML(
+                '<input type="submit" value="Opslaan" class="default" name="_save">',
+                str(response.content),
+            )
```
sergei-maertens, line 141: this is a very fragile way of asserting the behaviour, and moreover, it doesn't actually validate that the backend blocks the save, it just checks that a button is missing.

Better is to `form.submit(name="_save")` and assert that you get some kind of error response (I expect an HTTP 403, but it might be something else) *and* assert that the model instance was not updated.

src/woo_publications/metadata/tests/test_informatie_categorie_admin.py
```diff
@@ -0,0 +1,212 @@
...
+
+        response = self.app.get(url, user=self.user)
+        form = response.forms["informatiecategorie_form"]
+
+        self.assertEqual(response.status_code, 200)
+        # page has save button.
+        self.assertInHTML(
+            '<input type="submit" value="Opslaan" class="default" name="_save">',
+            str(response.content),
+        )
```
sergei-maertens, line 165: same remarks as earlier - use webtest for the actual form submitting and check what the result is of that

basically when using webtest, you're not supposed to be dealing with (raw) HTML yourself in your tests.

src/woo_publications/metadata/tests/test_informatie_categorie_admin.py
```diff
@@ -0,0 +1,212 @@
...
+        self.assertEqual(self.information_category_custom_entry.definitie, "changed")
+
+    def test_informatie_category_admin_create_item(self):
+        response = self.app.get(
+            reverse("admin:metadata_informatiecategorie_add"), user=self.user
+        )
+        form = response.forms["informatiecategorie_form"]
+
+        form["identifier"] = "https://www.example.com/waardenlijsten/999"
```
sergei-maertens, line 191: this is not something that users should be entering themselves, we need to generate it

src/woo_publications/metadata/tests/test_informatie_categorie_admin.py
```diff
@@ -0,0 +1,212 @@
...
+        form["naam"] = "new item"
+        form["naam_meervoud"] = "new items"
+        form["definitie"] = (
+            "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris risus nibh, "
+            "iaculis eu cursus sit amet, accumsan ac urna. Mauris interdum eleifend eros sed consectetur."
+        )
+
+        form.submit("submit")
```
sergei-maertens, line 199:
```suggestion
        form.submit(name="_save")
```

there's no submit button with attribute `name` set to `submit` - the available options are:

* `_save`
* `_addanother`
* `_continue`

See:

image

src/woo_publications/metadata/tests/test_informatie_categorie_admin.py
```diff
@@ -0,0 +1,212 @@
...
+                naam="new item",
+                naam_meervoud="new items",
+                definitie=(
+                    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris risus nibh, "
+                    "iaculis eu cursus sit amet, accumsan ac urna. Mauris interdum eleifend eros sed consectetur."
+                ),
+                oorsprong=InformatieCategorieOrigins.custom_entry,
+                order=2,
+            ).exists()
+        )
```
sergei-maertens, line 212: when this breaks, it's going to suck to figure out which aspect specifically is broken, so better is:

```py
added_item = InformationCategory.objects.order_by("-pk").first()
assert added_item is not None
self.assertEqual(added_item.naam, "new item")
self.assertTrue(added_item.identifier.startswith(CUSTOM_IDENTIFIER_URL_PREFIX))
self.assertEqual(added_item.naam_meervoud, ...)
...  # etc.
```
so when something fails, you know exactly which bit and can investigate that much more efficiently

src/woo_publications/metadata/admin.py
```diff
@@ -0,0 +1,29 @@
...
+        if obj and obj.oorsprong == InformatieCategorieOrigins.value_list:
+            return False
+        return True
+
+    def get_readonly_fields(self, request, obj=None):
+        readonly_fields = self.readonly_fields
+        if obj and obj.oorsprong == InformatieCategorieOrigins.custom_entry:
+            readonly_fields += ("identifier",)
+
+        return readonly_fields
```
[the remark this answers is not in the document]
bart-maykin: I was also thinking of adding the editable flag but I thought that we maybe wanted to prevent it so we can still change it from the terminal.

Also why do we need a default value? Is there a legit argument to add that? It seems weird to create none existing urls.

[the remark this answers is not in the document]
sergei-maertens: `editable=True` only controls the user interface, you can still change it through the ORM on the CLI.

> Also why do we need a default value? Is there a legit argument to add that? It seems weird to create none existing urls.

It's the cleanest way to achieve our desired behaviour - it serves both the API endpoints and admin at the same time at zero cost.

URLs do not necessarily need to be resolvable, they are often used for a unique namespace identifier. XML standards for example define namespaces as URLs that more often than not don't exist.

In this case, we only generate as URL to get a consistent shape of the data.

I'm suggesting a placeholder prefix now, in the future we could then easily find-and-replace that value if/when we have a more clever system and would actually allow resolving the URLs, but for now I want a fixed value so that we can easily move environments to different (test/acc/prod) domains.

sergei-maertens (approved):

PR #32, opened 2024-09-17
:sparkles: [#14] added list and retrieve api endpoints for informatie categorie

### LINKED ISSUE #14: ODRC: implement informatiecategorie resource

bart-maykin:
Partially Fixes #14

~~Depends on pr #31~~

**Changes**

created list and retrieve endpoints for informatie categorie.
updated oas schema
camelized queryparams


src/woo_publications/api/custom_filters.py
sergei-maertens, line 1: using `filters.py` instead of `custom_filters.py` is better - the `custom` is implied since otherwise we'd just import the thing :)

src/woo_publications/api/urls.py
sergei-maertens, line 1: Something went wrong in the rebase, the changes from 16cc87b55dc0618b9239ecb79320faa5ca5d9fb8 are gone

src/woo_publications/api/urls.py
```diff
@@ -4,16 +4,22 @@
...
 from rest_framework import routers
 
+from woo_publications.metadata.api.views.category import InformationCategoryViewset
+
 app_name = "api"
 
 router = routers.DefaultRouter()
+router.include_root_view = False
+
+router.register("informatie-categorie", InformationCategoryViewset)
```
sergei-maertens, line 14:
```suggestion
router.register("informatiecategorieen", InformationCategoryViewset)
```
it's a single word in Dutch, and for base/collection endpoints we use plural form

src/woo_publications/metadata/api/filterset/category.py
sergei-maertens, line 1: I don't think we need all the nesting - simply putting this in `src/woo_publications/metadata/api/filters.py` is enough :)

src/woo_publications/metadata/api/filterset/category.py
```diff
@@ -0,0 +1,28 @@
+from django.utils.translation import gettext_lazy as _
+
+from django_filters.rest_framework import FilterSet, filters
+
+from woo_publications.api.custom_filters import URLFilter
+from woo_publications.metadata.models import InformationCategory
```
sergei-maertens, line 6:
```suggestion
from ..models import InformationCategory
```
please use relative imports for names in the same django app

src/woo_publications/metadata/api/filterset/category.py
```diff
@@ -0,0 +1,28 @@
...
+        help_text=_(
+            "Search the information category based on the unique URI that identifies a specific category."
+        ),
+    )
+    naam = filters.CharFilter(
+        lookup_expr="exact",
+        help_text=_(
+            "Search the information category based on the name of the category."
+        ),
+    )
```
sergei-maertens, line 21: for auto-complete purposes this will need to support partial matches, so the `lookup_expr` should be `icontains`. Please add a test for this too.

src/woo_publications/metadata/api/serializer/category.py
sergei-maertens, line 1: Here too - please use a flatter structure, just `serializers.py` is enough. If files get too large, we break them up later.

src/woo_publications/metadata/api/serializer/category.py
```diff
@@ -0,0 +1,21 @@
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import serializers
+
+from woo_publications.metadata.models import InformationCategory
```
sergei-maertens, line 5: should be relative import

src/woo_publications/metadata/api/serializer/category.py
```diff
@@ -0,0 +1,21 @@
...
+        fields = (
+            "identifier",
+            "naam",
+            "naam_meervoud",
+            "definitie",
+            "oorsprong",
+            "order",
+        )
+        extra_kwargs = {
+            "order": {"help_text": _("The order number of the category.")},
```
sergei-maertens, line 20:
```suggestion
            "order": {"help_text": _("Controls the (default) ordering of categories in result lists.")},
```

I can see that it's an order number from the label and type information already, there's no need to mention that.

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
sergei-maertens, line 1: please put all the tests in `src/woo_publications/metadata/tests/` - stick to a flat structure.

Tests in subpackages hurt discoverability and always lead to test duplication and tests being smeared out in multiple places.

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
```diff
@@ -0,0 +1,152 @@
+from django.urls import reverse
+
+from rest_framework import status
+from rest_framework.test import APITestCase
+
+from woo_publications.metadata.constants import InformationCategoryOrigins
+from woo_publications.metadata.tests.factories import InformationCategoryFactory
```
sergei-maertens, line 7: relative imports

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class InformationCategoryTests(APITestCase):
+    @classmethod
+    def setUpTestData(self):
```
sergei-maertens, line 12: Please address these type checker warnings

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+            oorsprong=InformationCategoryOrigins.value_list,
+        )
+        self.information_category_custom_entry = InformationCategoryFactory.create(
+            identifier="https://www.example.com/waardenlijsten/2",
+            naam="second item",
+            naam_meervoud="second items",
+            definitie="This is some information about the second item.",
+            order=1,
+            oorsprong=InformationCategoryOrigins.custom_entry,
+        )
```
sergei-maertens, line 29: please set up the test data inside the test itself

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def test_list_informatie_categorie(self):
+        list_url = reverse("api:informationcategory-list")
+        response = self.client.get(list_url)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        data = response.json()
+        self.assertEqual(data["count"], 2)
```
sergei-maertens, line 37:
```suggestion
        list_url = reverse("api:informationcategory-list")

        response = self.client.get(list_url)

        self.assertEqual(response.status_code, status.HTTP_200_OK)
        data = response.json()
        self.assertEqual(data["count"], 2)
```
the visual separation between

```
setup
setup

act

assert
assert
```
is important to have well-readable tests :)

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # First item
```
sergei-maertens, line 39: consider using a `subTest` for each item - both of them will then run and a failure in one can still provide more information for the other one

src/woo_publications/metadata/api/tests/test_information_category_api_endpoints.py
```diff
@@ -0,0 +1,152 @@
...
+        self.assertEqual(data["results"][0]["naam"], "first item")
+        self.assertEqual(data["results"][0]["naamMeervoud"], "first items")
+        self.assertEqual(
+            data["results"][0]["definitie"],
+            "This is some information about the first item.",
+        )
+        self.assertEqual(data["results"][0]["order"], 0)
+        self.assertEqual(
+            data["results"][0]["oorsprong"], InformationCategoryOrigins.value_list
+        )
```
sergei-maertens, line 52: I prefer structuring these assertions in API tests as:

```py
expected = {
    "identifier": ...,
    "naam": ...,
    ...
}
self.assertEqual(data["results"][0], expected)
```

1. it's more readable what exactly is expected
2. if the assertion fails, the test itself will present a diff of which keys/values are not matching
3. it's easy to put a breakpoint for debugging to compare `expected` and `data["results"][0]` in a debug shell

src/woo_publications/metadata/api/views/category.py
sergei-maertens, line 1: This should just be `src/woo_publications/metadata/api/viewsets.py`

`viewsets.py` instead of `views.py` so that in the future if we have plain `APIView` subclasses, those go in `views.py`

src/woo_publications/metadata/api/views/category.py
```diff
@@ -0,0 +1,29 @@
+from django.utils.translation import gettext_lazy as _
+
+from drf_spectacular.utils import extend_schema, extend_schema_view
+from rest_framework import viewsets
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+
+from woo_publications.metadata.models import InformationCategory
```
sergei-maertens, line 7: relative import please :)

src/woo_publications/metadata/api/views/category.py
```diff
@@ -0,0 +1,29 @@
...
+        description=_("Alle informatie categorieën opvragen."),
+    ),
+    retrieve=extend_schema(
+        summary="Een specifiek informatie categorie opvragen.",
+        description="Een specifiek informatie categorie opvragen.",
+    ),
+)
+class InformationCategoryViewset(
+    ListModelMixin, RetrieveModelMixin, viewsets.GenericViewSet
+):
```
sergei-maertens, line 26:
```suggestion
class InformationCategoryViewSet(viewsets.ReadOnlyModelViewSet):
```

src/woo_publications/metadata/api/views/category.py
```diff
@@ -0,0 +1,29 @@
...
+from rest_framework import viewsets
+from rest_framework.mixins import ListModelMixin, RetrieveModelMixin
+
+from woo_publications.metadata.models import InformationCategory
+
+from ..filterset.category import InformationCategoryFilterset
+from ..serializer.category import InformationCategorySerializer
+
+
+@extend_schema(tags=["informatie categorie"])
```
sergei-maertens, line 13:
```suggestion
@extend_schema(tags=["informatiecategorie"])
```
correct spelling

src/woo_publications/metadata/api/views/category.py
```diff
@@ -0,0 +1,29 @@
...
+@extend_schema_view(
+    list=extend_schema(
+        summary=_("Alle informatie categorieën opvragen."),
+        description=_("Alle informatie categorieën opvragen."),
+    ),
+    retrieve=extend_schema(
+        summary="Een specifiek informatie categorie opvragen.",
+        description="Een specifiek informatie categorie opvragen.",
+    ),
```
sergei-maertens, line 22: Please do these in English, we will add translations later.

At any given time - marking a Dutch string as translatable is weird, the source language is English and we create a Dutch translation file to enter the Dutch strings. I don't want to introduce any inconsistency in those files.

src/woo_publications/metadata/api/views/category.py
```diff
@@ -0,0 +1,29 @@
...
+)
+class InformationCategoryViewset(
+    ListModelMixin, RetrieveModelMixin, viewsets.GenericViewSet
+):
+    queryset = InformationCategory.objects.all()
+    serializer_class = InformationCategorySerializer
+    filterset_class = InformationCategoryFilterset
```
sergei-maertens, line 29:
```suggestion
    queryset = InformationCategory.objects.all()
    serializer_class = InformationCategorySerializer
    filterset_class = InformationCategoryFilterset
    permission_classes = (permissions.AllowAny,)
```
can you make the permissions explicit for now? We'll add API keys later, but I prefer not having this implicitly defined.

src/woo_publications/metadata/models.py
```diff
@@ -20,7 +20,7 @@ class InformationCategory(OrderedModel):
     identifier = models.URLField(
         _("identifier"),
         help_text=_(
-            "The unique IRI that identifies this category in the overheid.nl value list. "
+            "The unique URI that identifies this category in the overheid.nl value list. "
```
sergei-maertens, line 23: Please revert this. An IRI is more specific than URI - it allows unicode charachters where a URI is limited to ASCII

src/woo_publications/utils/query.py
```diff
@@ -0,0 +1,20 @@
+from drf_spectacular.contrib.django_filters import DjangoFilterExtension
+
+from .utils import underscore_to_camel
+
+
+class CamelizeFilterExtension(DjangoFilterExtension):
```
sergei-maertens, line 6: this is API related and it's generic/base tooling, so it belongs in the `src/woo_publications/api/` app/package.

It's a good idea to create a `drf_spectacular` package in there, and then create `extensions.py` like in Open Forms: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/api/drf_spectacular/__init__.py

I think you can also just perform the imports in the `__init__.py` modules to register the extension, you don't need to use the `apps.py` for this.

src/woo_publications/utils/query.py
```diff
@@ -0,0 +1,20 @@
...
+    priority = 1
+
+    def get_schema_operation_parameters(self, auto_schema, *args, **kwargs):
+        """
+        camelize query parameters
+        """
+        parameters = super().get_schema_operation_parameters(
```
sergei-maertens, line 13: to address the coverage remarks, I'm okay with putting a `# pragma: no cover` on the class definition. We have CI checks that generate the API schema that ensure this behaves as expected, so we don't need unit tests for this.

src/woo_publications/utils/utils.py
```diff
@@ -0,0 +1,17 @@
+import re
+from typing import Union
```
sergei-maertens, line 2: we're using Python 3.12, this hasn't been required since Python 3.10 :)

```suggestion
```

src/woo_publications/utils/utils.py
sergei-maertens, line 1: this belongs in `src/woo_publications/api/utils.py`

src/woo_publications/utils/utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+from djangorestframework_camel_case.util import (
+    camelize_re,
+    underscore_to_camel as _underscore_to_camel,
+)
+
+
+def underscore_to_camel(input_: Union[str, int]) -> str | Union[str, int]:
```
sergei-maertens, line 10: with the new python 3.12 typing syntax, I think this should be:

```py
def underscore_to_camel[T: str | int](input_: T) -> T:
```
explanation:

1. the function takes a generic type `T`, which is either a `str` or an `int`
2. the function returns the same type as the input type (str -> str, and int -> int. str -> int or int -> str are not possible)
3. `str | int` is nicer syntax for `Union[str, int]`

`str | Union[str, int]` is also weird, since you're basically saying "it's `str` or `str` or `int`" -> the `str` union member is duplicated

src/woo_publications/utils/utils.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Convert a string from under_score to camelCase.
+    """
+    if not isinstance(input_, str):
+        return input_
```
sergei-maertens, line 15: please write some unit tests for this function, now there's no test coverage

sergei-maertens (changes requested):

src/woo_publications/api/drf_spectacular/extensions.py
```diff
@@ -0,0 +1,22 @@
...
+        camelize query parameters
+        """
+        parameters = super().get_schema_operation_parameters(  # pragma: no cover
+            auto_schema, *args, **kwargs
+        )
+
+        for parameter in parameters:
+            parameter["name"] = underscore_to_camel(
+                parameter["name"]
+            )  # pragma: no cover
```
sergei-maertens, line 20: See the annotation - you need to put the comment on line 17 or 18 for it to take effect (black formatting tends to trip these up)

src/woo_publications/api/tests/utils_tests.py
sergei-maertens, line 1: Test module name convention is `test_FOO.py`, not `utils_tests.py` - I'm not even sure these tests are properly discovered. So this file belongs in `src/woo_publications/api/tests/test_utils.py`

src/woo_publications/api/tests/utils_tests.py
```diff
@@ -0,0 +1,22 @@
+from unittest import TestCase
+
+from ..utils import underscore_to_camel
+
+
+class UnderscoreToCamelTests(TestCase):
+    def test_single_word_does_not_get_manipulated(self):
+        self.assertEqual(underscore_to_camel("single"), "single")
```
sergei-maertens, line 8: please stick to the visual blocks I've pointed out before:

```
setup

act

assert
```

So this would become:

```py
def test_single_word_does_not_get_manipulated(self):
    result = underscore_to_camel("single")

    self.assertEqual(result, "single")
```

this makes the tests so much more readable. When reading/debugging tests, at most I want to use 1% of my brain capacity to decipher the test code itself so that I have 99% to spend on The Actual Problem :tm:

src/woo_publications/api/urls.py
sergei-maertens: not resolved

src/woo_publications/api/urls.py
```diff
@@ -4,10 +4,14 @@
 from drf_spectacular.views import SpectacularJSONAPIView, SpectacularRedocView
 from rest_framework import routers
 
+from woo_publications.metadata.api.viewsets import InformationCategoryViewset
```
sergei-maertens, line 7:
```suggestion
from woo_publications.metadata.api.viewsets import InformationCategoryViewSet
```
I pointed this out in an earlier review - let's keep the casing consistent with the base classes

src/woo_publications/api/utils.py
```diff
@@ -0,0 +1,16 @@
...
+    underscore_to_camel as _underscore_to_camel,
+)
+
+
+def underscore_to_camel[T: str | int](input_: T) -> T:
+    """
+    Convert a string from under_score to camelCase.
+    """
+    if not isinstance(input_, str):
+        return input_
```
sergei-maertens, line 14: Tests indeed seem to not be picked up, since there are still missing coverage annotations.

src/woo_publications/metadata/api/filters.py
```diff
@@ -0,0 +1,29 @@
+from django.utils.translation import gettext_lazy as _
+
+from django_filters.rest_framework import FilterSet, filters
+
+from woo_publications.api.filters import URLFilter
+
+from ..models import InformationCategory
+
+
+class InformationCategoryFilterset(FilterSet):
```
sergei-maertens, line 10:
```suggestion
class InformationCategoryFilterSet(FilterSet):
```
please use consistent casing

src/woo_publications/metadata/api/filters.py
```diff
@@ -0,0 +1,29 @@
...
+    identifier = URLFilter(
+        lookup_expr="icontains",
+        help_text=_(
+            "Search the information category based on the unique IRI that identifies a specific category."
+        ),
+    )
```
sergei-maertens, line 16: I only remarked the partial match for the `naam`, the identifier should be an exact match. Nobody will every enter an identifier like this in a search box in the UI.

This should be `lookup_expr="exact"`

sergei-maertens (changes requested):
Almost there!

src/woo_publications/api/drf_spectacular/extensions.py
```diff
@@ -0,0 +1,20 @@
...
+        camelize query parameters
+        """
+        parameters = super().get_schema_operation_parameters(  # pragma: no cover
+            auto_schema, *args, **kwargs
+        )
+
+        for parameter in parameters:  # pragma: no cover
+            parameter["name"] = underscore_to_camel(parameter["name"])
+
+        return parameters  # pragma: no cover
```
sergei-maertens, line 20:
```suggestion
    def get_schema_operation_parameters(
        self, auto_schema, *args, **kwargs
    ):  # pragma: no cover
        """
        camelize query parameters
        """
        parameters = super().get_schema_operation_parameters(
            auto_schema, *args, **kwargs
        )

        for parameter in parameters:
            parameter["name"] = underscore_to_camel(parameter["name"])

        return parameters
```
can you try this? this should exclude the entire method from coverage

note that you can run coverage checks locally too:

```bash
coverage run src/manage.py test src
coverage html
# open htmlcov/index.html in the browser
```

src/woo_publications/api/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+        result = underscore_to_camel("single")
+
+        self.assertEqual(result, "single")
+
+    def test_multiple_words_get_manipulated(self):
+        result = underscore_to_camel("multiple_words_get_manipulated")
+        self.assertEqual(
+            result,
+            "multipleWordsGetManipulated",
+        )
```
sergei-maertens, line 17: I'm going to keep trying to hammer this in :smile: 

The visual separation with the blank line helps identifying the setup/act/assert blocks:

```suggestion
        result = underscore_to_camel("multiple_words_get_manipulated")

        self.assertEqual(
            result,
            "multipleWordsGetManipulated",
        )
```

sergei-maertens (approved):

PR #42, opened 2024-09-18
Add Documenten API to compose stack

### LINKED ISSUE #18: ODRC: add Documenten API to the stack
It was decided we will store the document data (and a bunch of metadata) in the existing VNG Documenten API. For local development/testing, we can use Open Zaak for this, for production usage this is also possible but other implementations exist.

This makes it possible for us to support large file uploads (chunks) out of the box, and we will also design the interaction from ODPC -> ODRC -> Documenten API around the large file uploads mechanism from the get-go.

The biggest challenge this poses, is implementing the Catalogi API parts in ODRC to be consumed by the Documenten API.

**Tasks**

- [ ] Add Open Zaak (pre-configured with Documenten API) to the docker compose stack
- [ ] Implement the `/catalogi/api/v1/informatieobjecttypen/:uuid` detail endpoint for each informatiecategorie exposed by the ODRC
    - include the properties `url`, `catalogus` (can be fixed/dummy), `omschrijving`, `vertrouwelijkheidaanduiding` (set to `openbaar`, by definition of WOO), `beginGeldigheid`, `informatieobjectcategorie` - these are the required properties
    - decide on authentication requirements

sergei-maertens:
Fixes #18

**Changes**

* Added Open Zaak to `docker-compose.yml`, with pre-configured fixtures
* Documented Open Zaak orchestration/setup
* Added system check to ensure we can consistently use the `host.docker.internal` hostname both in and outside of Docker
* Added `host.docker.internal` DNS entry to CI job to run the tests


src/woo_publications/contrib/documents_api/api.py
```diff
@@ -0,0 +1,78 @@
...
+    responses={200: CatalogiAPIDocumentTypeSerializer},
+)
+class CatalogiAPIDocumentTypeView(views.APIView):
+    permission_classes = (permissions.AllowAny,)
+
+    def get(self, request: Request, *args, **kwargs):
+        """
+        Expose an information category as a Catalogi API `informatieobjecttype`.
+        """
+        data = {
```
sergei-maertens, line 55: oops, need to add a proper unit test for this :)

src/woo_publications/contrib/documents_api/api.py
```diff
@@ -0,0 +1,78 @@
...
+    "vertrouwelijkheidaanduiding": VertrouwelijkheidsAanduidingen.openbaar,
+    "begin_geldigheid": "2024-09-01",
+    "concept": False,
+    "informatieobjectcategorie": "Placeholder",
+    "besluittypen": [],
+    "zaaktypen": [],
+}
+
+
+class CatalogiAPIDocumentTypeSerializer(serializers.Serializer):
```
bart-maykin, line 29: Can we add some help_texts? I know that this is temporary but it is still nice for the documentation.
And mark the required fields as such.

src/woo_publications/contrib/documents_api/tests/test_documents_api_integration.py
```diff
@@ -0,0 +1,63 @@
...
+            "creatiedatum": "2024-09-18",
+            "titel": "Test document",
+            "auteur": "Automated test suite",
+            "taal": "ENG",
+            "bestandsomvang": 0,
+        }
+
+        response = client.post("enkelvoudiginformatieobjecten", json=document_data)
+
+        self.assertEqual(response.status_code, 201, response.json())
```
bart-maykin, line 63: Can we make some checks to see if the data that gets returned is as expected?

src/woo_publications/contrib/documents_api/api.py
sergei-maertens: I'm only implementing the required fields, otherwise Open Zaak complains :) the default in DRF is `required=True`

src/woo_publications/contrib/documents_api/tests/test_documents_api_integration.py
sergei-maertens: I'll do that in another test so there's coverage for the endpoint. This is just a smoke test to validate that the docker compose services can actually poperly talk to each other (which was a birch to set up)

bart-maykin (approved):

PR #46, opened 2024-09-20
Feature/11 thema models

bart-maykin:
Partially Fixes #11

**Changes**

added theme model
added theme model factory

PR #47, opened 2024-09-20
Feature/11 admin

### LINKED ISSUE #11: ODRC: Toevoegen themas
**Description (why)**

**Specific details (how)**

**Test plan**

Theme properties:

* `identifier`: URI
* `name`: name of the theme

**Tasks**

    - List endpoint is paginated

bart-maykin:
Partially Fixes #11 

depends on #46 

**Changes**

Added treebeard
Added theme admin page
Added tests for theme admin


src/woo_publications/templates/admin/tree_change_list.html
bart-maykin, line 1: This file is made to alter the tree_change_list.html page within the treebeard library.
This template only gets used currently in conjuction with the Theme's models.
Which should only display the loaded in data from the government, so we want to restrict the user from alter the data.
Hence why I altered this file to remove the js files, which handles requests from the drag and drop feature.

sergei-maertens: can you add this comment to the template file itself too, for the future?

src/woo_publications/metadata/tests/test_theme_admin.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self.assertEqual(response.status_code, 403)
+
+    def test_theme_admin_can_not_create(
+        self,
+    ):
+        url = reverse("admin:metadata_theme_add")
+
+        response = self.app.get(url, user=self.user, expect_errors=True)
+        self.assertEqual(response.status_code, 403)
```
sergei-maertens, line 94:
```suggestion
        url = reverse("admin:metadata_theme_add")

        response = self.app.get(url, user=self.user, expect_errors=True)
        
        self.assertEqual(response.status_code, 403)
```

sergei-maertens (changes requested):
Can you add a django fixture for now (for demo purposes) in `src/woo_publications/fixtures/themes.json` with some of the themes from the waardelijst? 3 or so is enough, so that we can show something in the admin!

sergei-maertens (approved):

PR #53, opened 2024-09-24
Added sub theme fields to api spec

### LINKED ISSUE #11: ODRC: Toevoegen themas

bart-maykin:
Partially Fixes #11 

**Changes**

added cashed property to retrieve all the sub themes.
added nested view to display all sub themes on the theme api endpoints.


src/woo_publications/metadata/tests/test_theme_api_endpoints.py
```diff
@@ -62,7 +66,54 @@ def test_detail_theme(self):
...
+
+        with self.subTest("tree_dump_returns_empty_list"):
+            with patch(
+                "woo_publications.metadata.api.viewsets._normalize_tree_dump",
+                return_value=[{}],
+            ):
+
+                response = self.client.get(list_url)
+
+                self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
sergei-maertens, line 119: I don't understand why you'd break the implementation using unittest.mock and check the effect it has on the API endpoint.

If the implementation is broken, this should crash hard and report errors to Sentry, not "silently" fail with a 404 error.

Additionally, private functions (names defined with an underscore prefix) should *not* be tested in unit test, since they are private and an implementation detail by design. Tying tests and implementation details together removes the ability to refactor without having to touch tests.

sergei-maertens:
brain dump:

* use `LIKE ANY` with subquery (similar to Exists) to get optimized queryset
* use custom model iterable to add the children (flat structure into tree)
* this should then properly work with the built in queryset methods while still querying the children in the result set

sergei-maertens:
@bart-maykin I've looked at the implications of our earlier workarounds and possible "proper" solutions, and decided that taking the query performance hit for now is the quickest way to expose something. See #63 for the optimization ticket, which I hope we will get too but it might fall out of scope.

sergei-maertens (approved):

PR #59, opened 2024-09-30
Properly wire up the index/landing page

sergei-maertens:
Left-over from the default-project boilerplate, which I'll need to backport there too :)


sergei-maertens:
With Bart being on vacation I'm going to have to "self-approve" and merge this one :grimacing:

PR #67, opened 2024-10-09
Management command to create information category fixture

### LINKED ISSUE #14: ODRC: implement informatiecategorie resource

bart-maykin:
Fixes #14

**Changes**

Added func to create/update information categories.
Added mangagement command


src/woo_publications/metadata/management/commands/update_information_category.py
sergei-maertens, line 1: Let's rename this to `update_information_categories_from_waardenlijst` to be super explicit.

src/woo_publications/fixtures/informatie_category.json
```diff
@@ -0,0 +1,236 @@
+[
+{
+    "model": "metadata.informationcategory",
+    "pk": 1,
```
sergei-maertens, line 4: the fixture should be dumped with natural keys so that primary keys don't end up in the fixture, that *will* cause problems with existing instances and when the fixture gains new records over time

src/woo_publications/metadata/management/commands/update_information_category.py
```diff
@@ -0,0 +1,24 @@
+from django.core.management.base import BaseCommand, CommandError
+
+from woo_publications.metadata.update_informatie_category import (
+    InformatieCategoryWaardenlijstError,
+    update_informatie_category,
+)
+
+
+class Command(BaseCommand):
+    help = "Used to fetch the gov waardenlijsten data and turn them into a fixture to load the data into the db."
```
sergei-maertens, line 10:
```suggestion
    help = "Retrieve the information categories from the value list published on overheid.nl and dump them as a fixture."
```

src/woo_publications/metadata/management/commands/update_information_category.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--file-path",
+            action="store",
+            help="The file path to where the fixture file will be created.",
+            default=None,
```
sergei-maertens, line 17: I would specify the default value here in the command, and not in the function that implements it. The function itself can then be 100% agnostic about file paths.

Looking at the implementation of `dumpdata`, I think you can also pass `pathlib.Path` instances to `output` instead of converting it to a string first.

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+from pathlib import Path
+
+from django.conf import settings
+from django.core.management import call_command
+
+import requests
+from glom import PathAccessError, T, glom
+
+from woo_publications.metadata.constants import InformationCategoryOrigins
+from woo_publications.metadata.models import InformationCategory
```
sergei-maertens, line 11: these can be relative imports

```suggestion
from .constants import InformationCategoryOrigins
from .models import InformationCategory
```

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+
+SPEC = {
+    "naam": T["http://www.w3.org/2004/02/skos/core#prefLabel"][0]["@value"],
+    "naam_meervoud": T[
+        "https://identifier.overheid.nl/tooi/def/ont/prefLabelVoorGroepen"
+    ][0]["@value"],
+    "definitie": T["http://www.w3.org/2004/02/skos/core#definition"][0]["@value"],
+    "order": T["http://www.w3.org/ns/shacl#order"][0]["@value"],
+}
```
sergei-maertens, line 20: noice, neat and tidy!

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class InformatieCategoryWaardenlijstError(Exception):
+    pass
```
sergei-maertens, line 24: let's make the message attribute explicit

```suggestion
class InformatieCategoryWaardenlijstError(Exception):
    def __init__(self, message: str):
        self.message = message
        super().__init__(message)
```

so that in the management command you can access it: `raise CommandError(err.message) from err`

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def update_informatie_category(file_path: str):
```
sergei-maertens, line 27:
```suggestion
def update_informatie_category(file_path: Path):
```
and the management command is then responsible for ensure it's a `Path` instance that's passed:

```py
file_path = options["file_path"]
if not isinstance(file_path, Path):
    file_path = Path(file_path)
...
```

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+    response = requests.get(
+        "https://repository.officiele-overheidspublicaties.nl/waardelijsten/scw_woo_informatiecategorieen/3/json/scw_woo_informatiecategorieen_3.json"
+    )
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.ConnectionError as err:
+        raise InformatieCategoryWaardenlijstError(
+            "Could not connect with url."
+        ) from err
```
sergei-maertens, line 48: this will never raise ConnectionError`, since that would be raised already in the `requests.get` call.

Probably better to:

```py
try:
    response = requests.get(...)
except requests.RequestsError as err:
    raise InformatieCategoryWaardenlijstError("Could not retrieve the value list data.") from err

try:
    response.raise_for_status()
except requests.RequestError:
    raise InformatieCategoryWaardenlijstError(f"Got an unexpected response status code when retrieving the value list data: {response.status_code}.") from err

File: src/woo_publications/metadata/update_informatie_category.py [hunk 9]
```diff
@@ -0,0 +1,83 @@
...
+
+    data = response.json()
+    if not data:
+        raise InformatieCategoryWaardenlijstError("Could not retrieve json from url.")
```
sergei-maertens, line 52:
```suggestion
        raise InformatieCategoryWaardenlijstError("Received empty data from value list.")
```

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+        if (
+            not waardenlijst.get("@type")[0]
+            == "http://www.w3.org/2004/02/skos/core#Concept"
+        ):
+            continue
+
+        fields = glom(waardenlijst, SPEC, skip_exc=PathAccessError)
+        fields["oorsprong"] = InformationCategoryOrigins.value_list
+
+        if fields:
```
sergei-maertens, line 65: this will never be falsy because you assign `oorsprong` the line before

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+            InformationCategory.objects.update_or_create(
+                identifier=waardenlijst.get("@id"), defaults=fields
```
sergei-maertens, line 67: `.get` can return `None`, but `identifier` is never supposed to be `None`.

```suggestion
                identifier=waardenlijst["@id"], defaults=fields
```

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,83 @@
...
+    to_export = InformationCategory.objects.filter(
+        oorsprong=InformationCategoryOrigins.value_list
+    ).values_list("pk", flat=True)
+
+    call_command(
+        "dumpdata",
+        "metadata.informationcategory",
+        format="json",
+        indent=4,
+        natural_primary=True,
```
sergei-maertens, line 79: you'll need to do some implementation work to make this actually work properly in the fixture :)

sergei-maertens (changes requested):

src/woo_publications/metadata/management/commands/update_information_categories_from_waardenlijst.py
```diff
@@ -0,0 +1,37 @@
+from pathlib import Path
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+from woo_publications.metadata.update_informatie_category import (
```
sergei-maertens, line 6:
```suggestion
from ...update_informatie_category import (
```
please use relative paths within the same django app

src/woo_publications/metadata/management/commands/update_information_categories_from_waardenlijst.py
```diff
@@ -0,0 +1,37 @@
...
+    InformatieCategoryWaardenlijstError,
+    update_informatie_category,
```
sergei-maertens, line 8:
```suggestion
    update_information_category,
```
weird NL/EN mix in the code, let's try to keep code as much English as possible

src/woo_publications/metadata/management/commands/update_information_categories_from_waardenlijst.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--file-path",
+            action="store",
+            help="The file path to where the fixture file will be created.",
+            default=Path(
+                settings.BASE_DIR
+                / "src"
+                / "woo_publications"
```
sergei-maertens, line 23:
```suggestion
                settings.DJANGO_PROJECT_DIR
```

src/woo_publications/metadata/management/commands/update_information_categories_from_waardenlijst.py
```diff
@@ -0,0 +1,37 @@
...
+                / "fixtures"
+                / "informatie_category.json"
```
sergei-maertens, line 25:
```suggestion
                / "information_categories.json"
```

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,82 @@
...
+    data = response.json()
+    if not data:
+        raise InformatieCategoryWaardenlijstError(
+            "Received empty data from value list."
+        )
+
+    for waardenlijst in data:
+        # filter out all ids that aren't waardenlijsten
+        if (
+            not waardenlijst.get("@type")[0]
```
sergei-maertens, line 56: `.get(...)` can return `None`, so you'd get a type error here because `None[0]` doesn't work.

```suggestion
            not waardenlijst["@type"][0]
```

src/woo_publications/metadata/update_informatie_category.py
sergei-maertens: I don't understand why we need this check - the previous version never failed/crashed if we didn't perform the check, so why do we need it?

src/woo_publications/metadata/update_informatie_category.py
```diff
@@ -0,0 +1,82 @@
...
+        ):
+            continue
+
+        fields = glom(waardenlijst, SPEC, skip_exc=PathAccessError)
+
+        if fields:
+            fields["oorsprong"] = InformationCategoryOrigins.value_list
+            InformationCategory.objects.update_or_create(
+                identifier=waardenlijst["@id"], defaults=fields
+            )
```
sergei-maertens, line 67:
```suggestion
            fields["oorsprong"] = InformationCategoryOrigins.value_list
            InformationCategory.objects.update_or_create(
                identifier=waardenlijst["@id"], 
                defaults={**fields, "oorsprong": InformationCategoryOrigins.value_list},
            )
```
this way you can avoid mutating the dict - it's a good practice to apply the functional programming principle that mutations should be avoided

src/woo_publications/metadata/update_informatie_category.py
sergei-maertens, line 1: I think `waardenlijst_sync.py` is a better module name to capture what's happening in this file.

src/woo_publications/metadata/tests/test_waardenlijst_sync.py
```diff
@@ -0,0 +1,84 @@
...
+            file_path = Path(file.name)
+
+            update_information_category(file_path)
+
+            with self.subTest("database populated"):
+                self.assertEqual(InformationCategory.objects.count(), 18)
+
+            with self.subTest("user content does not cause conflicts"):
+                information_category = InformationCategory.objects.order_by("pk").last()
+                information_category.identifier = "https://something.else"
```
sergei-maertens, line 34: adding `assert information_category is not None` will fix the type checker issue :)

src/woo_publications/metadata/tests/test_waardenlijst_sync.py
```diff
@@ -0,0 +1,84 @@
...
+from pathlib import Path
+from uuid import uuid4
+
+from django.core.management import call_command
+from django.test import TestCase
+
+import requests
+import requests_mock
+
+from ...utils.tests.vcr import VCRMixin
```
sergei-maertens, line 12:
```suggestion
from woo_publications.utils.tests.vcr import VCRMixin
```

sergei-maertens (approved):

PR #72, opened 2024-10-10
Theme management command

### LINKED ISSUE #11: ODRC: Toevoegen themas

bart-maykin:
Fixes #11 

**Changes**

Added magagement command to store theme waardenlijsten data and turn it into a fixture


bart-maykin:
I renamed the `waardenlijst_sync.py` file to `information_category_sync.py` to make these sync files more readable. Is this a good decision or should we not split the code up in different files? If we go this route then it might also be an idea to create a dir to store them in a better place.

sergei-maertens:
> I renamed the `waardenlijst_sync.py` file to `information_category_sync.py` to make these sync files more readable. Is this a good decision or should we not split the code up in different files? If we go this route then it might also be an idea to create a dir to store them in a better place.

separate files is great, once we have all three management commands and implementations, we can see if abstractions/refactors/restructuring are in order or not :)

src/woo_publications/metadata/theme_sync.py
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+class ThemeWaardenlijstError(Exception):
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(message)
+
+
+def update_theme(file_path: Path):
```
sergei-maertens, line 29: `file_path` here is annotated as a `glom.Path`, not a `pathlib.Path` which is causing you type errors.

Removing `from glom import Path` and replacing it with `from pathlib import Path` should fix all issues

bart-maykin: Yes just noticed it myself...

bart-maykin:
> > I renamed the `waardenlijst_sync.py` file to `information_category_sync.py` to make these sync files more readable. Is this a good decision or should we not split the code up in different files? If we go this route then it might also be an idea to create a dir to store them in a better place.
> 
> separate files is great, once we have all three management commands and implementations, we can see if abstractions/refactors/restructuring are in order or not :)

Cool then we are on the same page. I was looking into making the code better because there is a lot of duplicate code. But then I remembered that we are gonna have another command so I left it as it is.

src/woo_publications/fixtures/themes.json
sergei-maertens, line 1: self-reminder: ensure this fixture is loaded on container startup

src/woo_publications/metadata/theme_sync.py
```diff
@@ -0,0 +1,85 @@
...
+    ]
+
+    root_themes_data = [theme["data"] for theme in waardenlijst if not theme["parents"]]
+    sub_themes = [theme for theme in waardenlijst if theme["parents"]]
+
+    # create/update root themes
+    for theme in root_themes_data:
+        theme_object = Theme.objects.filter(identifier=theme["identifier"])
+        if theme_object.exists():
+            theme_object.update(**theme)
```
sergei-maertens, line 60: can you address the missing test coverage?

src/woo_publications/metadata/theme_sync.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            Theme.add_root(**theme)
+
+    # create/update sub themes
+    for theme in sub_themes:
+        theme_object = Theme.objects.filter(identifier=theme["data"]["identifier"])
+        for parent in theme["parents"]:
+            if theme_object.exists():
+                theme_object.update(**theme["data"])
```
sergei-maertens, line 69: this wouldn't take into account if the parent changes, would it?

sergei-maertens (changes requested):

sergei-maertens (approved):
Nice one!

PR #101, opened 2024-10-25
Token authorization and api permissions

### LINKED ISSUE #17: ODCR: set up API keys and API Authn/Authz model
The publication component and citizen portal will consume the registration API. For this, they need to authenticate using an API key.

**Tasks**

- ~~Define two roles: `publication_component`, `citizen_portal`~~ SCRAPPED
    - publication component implies read + write of all data
    - citizen portal implies read of published (!) data only
- [x] Define data model + set up admin for API Key, similar to objects API
    - API key, (securily) randomly generated string
    - Contact person name, phone + email address (optional fields)
    - Roles to apply, which is one or more of the two roles above (simple `ArrayField` should suffice)
- [x] Define DRF authentication class
    - Authenticate based on the API key in the request header `Authorization: Token <key>`
    - This is the only allowed authentication class (so don't expose/enable BasicAuth, TokenAuth, SessionAuth...)
    - Must be properly documented in the generated API schema
- [x] Define permission classes for read/write permissions, which are inferred/taken from the roles associated with the API key

bart-maykin:
Fixes #17

**Changes**

Added api token with permissions.
Added api token authorization to api endpoints.


src/woo_publications/fixtures/default_admin_index.json
sergei-maertens, line 1: where's the entry for the new model?

Now it's showing up in "overige"

image

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,69 @@
...
+        blank=True,
+    )
+    last_modified = models.DateTimeField(
+        _("last modified"),
+        auto_now=True,
+        help_text=_("Last date when the token was modified"),
+    )
+    created = models.DateTimeField(
+        _("created"), auto_now_add=True, help_text=_("Date when the token was created")
+    )
```
sergei-maertens, line 57: we're not exposing this in any API endpoints, so please define the model fields in English.

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,69 @@
...
+        "choices": [
+            {"title": label, "value": value}
+            for value, label in PermissionOptions.choices
+        ],
+        "widget": "multiselect",
+    },
+}
+
+
+class TokenAuth(models.Model):
```
sergei-maertens, line 25: I thought we had discussed that the model name should be `Application`? You created an app `application` then and I told you to put the model in `api.models`, but it should still be named `Application` because that's what it represents.

This keeps terminology close to the OAUTH application concept and should make the admin interface easier to use.

For the admin index menu, I think it makes sense to create a group "API access" containing this model.

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,69 @@
...
+    token = models.CharField(_("token"), max_length=40, primary_key=True)
```
sergei-maertens, line 26: You cannot use this as a primary key because then it shows up in the URLs in the admin interface, and thus in the server access logs.

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,69 @@
...
+    permissies = JSONField(
+        _("permission"),
+        help_text=_("The permission this API token has access to."),
+        schema=PERMISSION_SCHEMA,
+        blank=True,
+        null=True,
+    )
```
sergei-maertens, line 33: `JSONField` is much too wide, this really should be an array field so we get proper database constraints to guarantee the structure of our data:

```py
PERMISSION_SCHEMA = {
    "type": "array",
    "items": {
        "type": "string",
        "choices": [
            {"title": label, "value": value}
            for value, label in PermissionOptions.choices
        ],
    },
}

...
    permissies = ArrayField(
        models.CharField(max_length=20, choices=PermissionOptions.choices),
        verbose_name=_("permission"),
        help_text=_("The permission this API token has access to."),
        schema=PERMISSION_SCHEMA,
        blank=True,
        default=list,
    )
```
this uses the `ArrayField` from `django-jsonform` 2.23+.

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,69 @@
...
+    contact_persoon = models.CharField(
+        _("contact person"),
+        max_length=200,
+        help_text=_("Name of the person who can access the API"),
```
sergei-maertens, line 37: I don't think these help texts are very useful. It's the contact person, but what they do with the API keys after that is up to them.

"Name of the person to contact about this application and the associated credentials." is probably better phrasing. The same remark applies to the other fields.

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,69 @@
...
+        auto_now=True,
+        help_text=_("Last date when the token was modified"),
+    )
+    created = models.DateTimeField(
+        _("created"), auto_now_add=True, help_text=_("Date when the token was created")
+    )
+
+    class Meta:
+        verbose_name = _("token authorization")
+        verbose_name_plural = _("token authorizations")
```
sergei-maertens, line 61:
```suggestion
    class Meta:
        verbose_name = _("application API key")
        verbose_name_plural = _("application API keys")
```
probably best to display it like that in the UI.

src/woo_publications/api/tests/factories.py
```diff
@@ -0,0 +1,21 @@
...
+
+    class Params:
+        read_permission = factory.Trait(permissies=[PermissionOptions.read.value])
+        write_permission = factory.Trait(permissies=[PermissionOptions.write.value])
+        read_write_permission = factory.Trait(
+            permissies=[
+                PermissionOptions.read.value,
+                PermissionOptions.write.value,
+            ]
+        )
```
sergei-maertens, line 21: you should be able to drop the `.value` everywhere, it's not needed.

src/woo_publications/api/tests/mixins.py
```diff
@@ -0,0 +1,181 @@
...
+AUDIT_HEADERS = {
+    "AUDIT_USER_REPRESENTATION": "username",
+    "AUDIT_USER_ID": "id",
+    "AUDIT_REMARKS": "remark",
+}
+
+
+class APIKeyUnAuthorizedMixin:
+
+    self: APITestCase
```
sergei-maertens, line 15: thought I had on Friday evening: this does nothing, as it would refer to `self.self`, which doesn't make any sense. So I'm guessing removing this line will change nothing.

src/woo_publications/api/tests/mixins.py
```diff
@@ -0,0 +1,181 @@
...
+    client: APIClient
+
+    def assertWrongApiKeyProhibitsGetEndpointAccess(self, url):
+        no_permission_token = TokenAuthFactory.create().token
+        write_token = TokenAuthFactory.create(write_permission=True).token
+
+        with self.subTest("no token given"):  # type: ignore reportAttributeAccessIssue
+            response = self.client.get(url, headers=AUDIT_HEADERS)
+            assert response.status_code == status.HTTP_401_UNAUTHORIZED
```
sergei-maertens, line 24: bare asserts is pytest-style, but we're doing unittest style, so this really needs to be

```suggestion
            self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)  # type: ignore reportAttributeAccessIssue
```
there's a reason these "wrappers" exist - it's so they can provide extra context and print appropriate messages to signal test failure.

src/woo_publications/api/tests/mixins.py
```diff
@@ -0,0 +1,181 @@
...
+}
+
+
+class APIKeyUnAuthorizedMixin:
+
+    self: APITestCase
+    client: APIClient
+
+    def assertWrongApiKeyProhibitsGetEndpointAccess(self, url):
+        no_permission_token = TokenAuthFactory.create().token
```
sergei-maertens, line 19:
```suggestion
        no_permission_token = TokenAuthFactory.create(permissions=[]).token
```
do not rely on implicit factory data for explicit test cases, set up your data explicitily to leave no doubt

src/woo_publications/api/tests/mixins.py
```diff
@@ -0,0 +1,181 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()  # type: ignore reportAttributeAccessIssue
+
+        cls.token_auth = TokenAuthFactory.create(read_write_permission=True)
+
+    def setUp(self):
+        super().setUp()  # type: ignore reportAttributeAccessIssue
+
+        self.client.credentials(HTTP_AUTHORIZATION=f"Token {self.token_auth.token}")
```
sergei-maertens, line 181: This is proper use of the `setUp` method, nice :+1:

sergei-maertens (changes requested):

src/woo_publications/fixtures/default_admin_index.json
bart-maykin: I ran the `generate_admin_index_fixture.sh` file and it didn't generate it. I will add the `organisations` and new model manually and look into this issue in a new ticket.

src/woo_publications/api/tests/mixins.py
bart-maykin: I did it this way to reduce the amount of `type: ignore reportAttributeAccessIssue` that was used in the file. But changed it based on your feedback.

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,76 @@
...
+
+PERMISSION_SCHEMA = {
+    "type": "list",
+    "items": {
+        "type": "string",
+        "choices": [
+            {"title": label, "value": value}
+            for value, label in PermissionOptions.choices
+        ],
+        "widget": "multiselect",
```
sergei-maertens, line 20:
```suggestion
```
not needed - my sample code excluded it so that the UI made it clear you're dealing with a list of permissions.

image

Does not look great tbh

src/woo_publications/api/models.py
```diff
@@ -0,0 +1,76 @@
...
+        auto_now=True,
+        help_text=_("Last date when the token was modified"),
+    )
+    created = models.DateTimeField(
+        _("created"), auto_now_add=True, help_text=_("Date when the token was created")
+    )
+
+    class Meta:
+        verbose_name = _("application API key")
+        verbose_name_plural = _("application API key")
```
sergei-maertens, line 68:
```suggestion
        verbose_name_plural = _("application API keys")
```

src/woo_publications/api/authorization.py
```diff
@@ -0,0 +1,16 @@
+from django.utils.translation import gettext_lazy as _
+
+from rest_framework import exceptions
+from rest_framework.authentication import TokenAuthentication as _TokenAuthentication
+
+
+class TokenAuthentication(_TokenAuthentication):
+    def authenticate_credentials(self, key):
+        from .models import Application
```
sergei-maertens, line 9: I moved the import to the top of the module and things continue to work - why is it a local import?

src/woo_publications/api/permissions.py
```diff
@@ -17,3 +20,28 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+
+class TokenAuthPermission(BasePermission):
+    def has_permission(self, request, view) -> bool:
+        token = request.auth
+
+        if not isinstance(token, Application):
+            return False
+
+        if not isinstance(token.permissions, list):
+            return False
```
sergei-maertens, line 33: this will always be a list, because the model defines it as an `ArrayField`. If this line is here to satisfy the type checker, then do:

```py
assert isinstance(token.permissions, list)
```

now there is missing branch coverage for a situation that cannot happen

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #103, opened 2024-10-30
Publication api endpoint owner field

### LINKED ISSUE #69: Expose owner/creator of Publicatie and allow filtering
### Thema / Theme

API

### Omschrijving / Description

Filtering the "own" publications in the UI (by the ODPC BFF) requires the filter parameter to exist on our end, and the information to be extracted.

This depends on the #16 and audit logging added via #35 to register the information about "ownership".

Extra field to expose on the `Publicatie` resource:

```yaml
eigenaar:
  type: object
  properties:
    weergaveNaam:  # obtained from the audit trail request headers, extracted from the audit trails/log entries
      type: string
    identifier:  # arbitrary, unique-ish string. Could be a `sub` claim from OIDC, but for our perspective this data has no "meaning"
      type: string
```

**Tasks**

- [x] Expose `eigenaar` field
- [x] Add filter parameter `eigenaar` to list API endpoint for `Publicatie`. The field must be an exact match on `eigenaar.identifier`.

### Toegevoegde waarde / Added value

Being able to finish GeneriekPublicatiePlatformWoo/Openbaar-Documenten-Publicatie-Component#8

### Aanvullende opmerkingen / Additional context

_No response_

bart-maykin:
Fixes #69

**Changes**

- Added serializer field to retrieve owner instance from timelinelogger
- Added filter field to filter on owner.identifier

### DIFF
--- a/src/woo_publications/publications/api/filters.py
+++ b/src/woo_publications/publications/api/filters.py
@@ -62,10 +62,7 @@ class PublicationFilterSet(FilterSet):
         )
 
     def filter_eigenaar(self, queryset, name: str, value: str):
-        publication_ct = ContentType.objects.get(
-            app_label="publications",
-            model="publication",
-        )
+        publication_ct = ContentType.objects.get_for_model(Publication)
 
         publication_object_ids = TimelineLogProxy.objects.filter(
             content_type=publication_ct,
--- a/src/woo_publications/publications/api/filters.py
+++ b/src/woo_publications/publications/api/filters.py
@@ -1,6 +1,10 @@
+from django.contrib.contenttypes.models import ContentType
 from django.utils.translation import gettext_lazy as _
 
 from django_filters.rest_framework import FilterSet, filters
+
+from woo_publications.logging.constants import Events
+from woo_publications.logging.models import TimelineLogProxy
 
 from ..models import Document, Publication
 
@@ -45,7 +49,28 @@ class PublicationFilterSet(FilterSet):
             "verkorte_titel",
         ),
     )
+    eigenaar = filters.CharFilter(
+        help_text=_("Filter publications based on the owner identifier of the object."),
+        method="filter_eigenaar",
+    )
 
     class Meta:
         model = Publication
-        fields = ("sorteer",)
+        fields = (
+            "sorteer",
+            "eigenaar",
+        )
+
+    def filter_eigenaar(self, queryset, name: str, value: str):
+        publication_ct = ContentType.objects.get(
+            app_label="publications",
+            model="publication",
+        )
+
+        publication_object_ids = TimelineLogProxy.objects.filter(
+            content_type=publication_ct,
+            extra_data__event=Events.create,
+            extra_data__acting_user__identifier=value,
+        ).values_list("object_id", flat=True)
+
+        return queryset.filter(pk__in=[id for id in publication_object_ids])
--- a/src/woo_publications/publications/api/serializer.py
+++ b/src/woo_publications/publications/api/serializer.py
@@ -1,10 +1,6 @@
 from django.utils.translation import gettext_lazy as _
 
-from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
-
-from woo_publications.logging.constants import Events
-from woo_publications.logging.models import TimelineLogProxy
 
 from ..models import Document, Publication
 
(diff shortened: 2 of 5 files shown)


src/woo_publications/publications/api/serializer.py
```diff
@@ -29,14 +33,34 @@ class Meta:  # type: ignore
...
+        help_text=_(
+            "arbitrary, unique-ish string. Could be a `sub` claim from OIDC, but for our perspective this data has no 'meaning'."
+        ),
+    )
+
+
 class PublicationSerializer(serializers.ModelSerializer):
+    eigenaar = serializers.SerializerMethodField(
+        help_text=_("De eigenaar van de publicatie."), method_name="get_acting_user"
+    )
```
sergei-maertens, line 54: `SerializerMethodField` is something I hope you forget that it exists - it's really a last resort to get the data out.

The reason is that it doesn't play nice with the generated API documentation - evidenced by your need for `extend_schema_field`, and that is a risk of implementation and documentation diverging.

Ultimately serialization is about "I have rich python/django objects, translate that into something that I can chuck over the wire and put in JSON". Serializers themselves should really be as boring as possible.

So, what is the correct approach then?

1. Move the lookups/data retrieval into their appropriate layer - either the view or the model layer. In this case, the model layer is easiest.
2. Define the serializer properly "inline" and mark it as read-only

```suggestion
    eigenaar = EigenaarSerializer(
        source="get_owner",
        label=_("owner"),
        help_text=_("The creator of the publication, derived from the audit logs."),
        allow_null=True,
        read_only=True,
    )
```

Notable changes:

- I made everything English
- I use the `EigenaarSerializer` directly - it knows how to serialize the `ActingUser` data
- I added `allow_null=True` - rather than returning an empty object if we don't have the information, let's just return `null` for that, as that holds the right semantic value. The client then knows if it's not None, that it will  be an object with the keys `displayName` and `identifier`
- I specified `source="get_owner"`, which is a method on the model.

`PublicationSerializer` gets a `Publication` instance, so when it tries to serialize the `eigenaar`, it will look up the `get_owner` attribute, see that it's a method that can be called, and call it. That returns the dictionary with information, or `None` - if it's the latter, it won't even attempt to use `EigenaarSerializer`, but will stick with `None`. If we have the data, the return value of `get_object` is passed to `EigenaarSerializer` and we get the nested object.

The `get_owner` implementation that works is pretty much what you have in `get_acting_user`:

```py
    def get_owner(self) -> ActingUser | None:
        """
        Extract the owner from the audit trails.
        """
        qs = TimelineLogProxy.objects.for_object(self)  # type: ignore reportAttributeAccessIssue
        try:
            log = qs.get(extra_data__event=Events.create)
        except TimelineLogProxy.DoesNotExist:
            return None
        assert isinstance(log, TimelineLogProxy)
        return log.acting_user[0]
```
with some proper type hints/assertions - I'll have to take a look at the library timeline-logger library :)

src/woo_publications/publications/api/filters.py
```diff
@@ -45,7 +49,28 @@ class PublicationFilterSet(FilterSet):
...
+        fields = (
+            "sorteer",
+            "eigenaar",
+        )
+
+    def filter_eigenaar(self, queryset, name: str, value: str):
+        publication_ct = ContentType.objects.get(
+            app_label="publications",
+            model="publication",
+        )
```
sergei-maertens, line 68:
```suggestion
        publication_ct = ContentType.objects.get_for_model(Publication)
```

src/woo_publications/publications/api/filters.py
```diff
@@ -45,7 +49,28 @@ class PublicationFilterSet(FilterSet):
...
+
+        publication_object_ids = TimelineLogProxy.objects.filter(
+            content_type=publication_ct,
+            extra_data__event=Events.create,
+            extra_data__acting_user__identifier=value,
+        ).values_list("object_id", flat=True)
```
sergei-maertens, line 74: I'll allow this for now, but I am a little bit concerned about the model being used directly rather than through helper functions. The `extra_data` internal structure is really something that should be contained within `woo_publications.logging`. and not leak outside. E.g. a nicer Python API would be:

```py
from woo_publications.logging.service import filter_logs 

logs = filter_logs(content_type=publication_ct, event=Events.create, acting_user=value)
publication_object_ids = logs.values_list("object_id", flat=True)
```

no need to make this change now, let's first see if this approach survives first contact :sweat_smile:

src/woo_publications/publications/api/filters.py
```diff
@@ -45,7 +49,28 @@ class PublicationFilterSet(FilterSet):
...
+
+        return queryset.filter(pk__in=[id for id in publication_object_ids])
```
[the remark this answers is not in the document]
bart-maykin: image

It returns this error if you do it directly with `publication_object_ids`.
That is why I transformed it into a list in the filter.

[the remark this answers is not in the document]
sergei-maertens: I see, I had hoped Django/Postgres was smarter and able to do the casting itself.

src/woo_publications/publications/api/serializer.py
```diff
@@ -29,14 +29,38 @@ class Meta:  # type: ignore
         )
 
 
+class EigenaarSerializer(serializers.Serializer):
+    display_name = serializers.CharField(
+        read_only=True,
+        help_text=_(
+            "obtained from the audit trail request headers, extracted from the audit trails/log entries."
+        ),
+    )
```
sergei-maertens, line 38:
```suggestion
    weergave_naam = serializers.CharField(
        source="display_name",
        read_only=True,
        help_text=_("The display name of the user, as recorded in the audit trails."),
    )
```
per the ticket specification, using Dutch attribute names in the API documentation.

src/woo_publications/publications/api/serializer.py
```diff
@@ -29,14 +29,38 @@ class Meta:  # type: ignore
...
+    identifier = serializers.CharField(
+        read_only=True,
+        help_text=_(
+            "arbitrary, unique-ish string. Could be a `sub` claim from OIDC, but for our perspective this data has no 'meaning'."
+        ),
```
sergei-maertens, line 43: the comments in the ticket were meant as a hint for you, the developer, not as the help text of the field :stuck_out_tongue: 

```suggestion
        help_text=_("The system identifier that uniquely identifies the user performing the action."),
```
text taken from `src/woo_publications/api/drf_spectacular/headers.py`

PR #106, opened 2024-10-30
Document creation and upload - implement Documents API client

### LINKED ISSUE #19: ODRC: implement Document creation/upload
**Acceptance criteria**

Copied/moved from #2

- Het format wordt gevalideerd wanneer het bestand geüpload wordt.
- Via een API kan een document (data-object) gecreëerd, geraadpleegd, gemuteerd en verwijderd worden.
- Bij een document moet exact één bestand (bijvoorbeeld een PDF) worden opgeslagen.
- De creatie, mutatie of verwijdering van een document wordt gelogd (zie #16)

## Document creation + uploads

We make use of the file parts mechanism of the Documenten API, always.

* ODRC will proxy to the underlying documenten API

### Preparing a document upload

The client must pass the necessary metadata and then based on the response of that,
cut up the file upload in parts that can be submitted individually.

The request body schema of a Document POST would look something like:

```yaml
Document:
  type: object
  required:
    - identifier
  properties:
    publicatie:  # UUID of the publication it belongs to
      type: string
      format: uuid
    identifier:  # the 'primary' identifier
      type: string
    creatiedatum:
      type: string
      format: date
    officieleTitel:  # DiWoo doesn't seem to apply a max length
      type: string
    verkorteTitel:
      type: string
    omschrijving:
      type: string
    # from waardelijst -> expose options in separate endpoint (!)
    # POST (write) operations should be able to just provide the identifier IRI instead of this complex object if ICATT desires this
    bestandsformaat:  
      type: object
      properties:
        identifier:  # IRI from waardelijst
          type: string
          format: uri
        mimeType:  # e.g. application/pdf
          type: string
        naam:
          type: string  # e.g. "PDF"
```

This translates to a request of ODRC -> Documenten API with schema:

```yaml
EnkelvoudigInformatieobject:
  type: object
  properties:
    identificatie:  # primary identifier of Document OR let the Documenten API generate one?
      type: string
    bronorganisatie:  # fixed, global configuration parameter in ODRC initially, could become 'smart' in the future
      type: string
    creatiedatum:  # taken from Document.creatiedatum
      type: string
      format: date
    titel:
      type: string
      minLength: 1
      maxLength: 200
    auteur:  # taken from Document.publicatie.organization
      type: string
    status:
      const: definitief  # archiving will move this to gearchiveerd, later
    formaat:  # derived from Document.bestandsformaat
      type: string
    taal:  # derived from Document.taal -> convert to/from ISO 639-2/B
      type: string
      enum:
        - dut
        - eng
    bestandsnaam:  # taken from Document.bestandsnaam
      type: string
    bestandsomvang:  # taken from Document.bestandsomvang, prepares the file parts
      type: number
    indicatieGebruiksrecht:
      const: false
    informatieobjecttype:  # points to /catalogi/api/v1/informatieobjecttypen/:uuid for the Document.publicatie.informatiecategorie
      type: string
      format: uri
```

The Documenten API will return a `lock` and list of BestandsDelen for upload, each
bestandsdeel will have the shape:

```yaml
BestandsDeel:
  type: object
  properties:
    url:  # URL to PUT to
      type: string
      format: uri
    volgnummer:
      type: integer
    omvang:
      type: integer
    voltooid:
      const: false
    lock: # ??
      type: string
```

The ODRC will then expose endpoints for these part uploads so the publication component
can upload the parts:

URL: `/api/v1/documenten/:uuid/bestandsdeel/:index`

A bestandsdeel wil simply be multipart/form-data, with the API key as auth header.

The request will be transformed by the ODRC, which adds the lock ID & JWT for the 
Documenten API, and streams the file part down to the Documenten API.

Once all parts are received, we unlock the created document.

**Tasks**

- [x] Implement API endpoint/serializer for `POST /api/v1/documenten`
    - Part of the metadata is stored in our database
    - Other metadata that we can store in Documenten API, we store there
    - Record the requested bestandsdelen from the Documenten API
    - Record the lock ID to finalize the upload/creation
    - Resolve the internal "Catalogi API" endpoint for the informatieobjecttype
- [x] Implement the API endpoint/serializer for `PUT /api/v1/documenten/:uuid/bestandsdelen/:index`
    - Use the stored lock ID & other configuration/metadata to all the Documenten API `/api/v1/bestandsdelen/:uuid`
    - If all parts are completely uploaded, unlock the document
    - Emit in the response body `documentFinalized: true|false` so that the client can be informed that they can refresh the document resource if needed, as file uploads will likely happen in parallel
    - Nice to have: check if we can make use of proxying/streaming (timebox: 1 day)
- ~~The first part of the document contains the magic bytes that can be used to validate against the `format` in the metadata. Check the upload validation in Open Forms for inspiration (and edge cases).~~ postponed to #36

sergei-maertens:
Partly fixes #19

**Changes**

This implements the low-level HTTP client to talk to the Documents API. These clients will be instantiated and used in our API endpoints.

I've added tests for the expected flows on how this will be called. The idea is that:

1. ODPC makes API call to us to create a document, and tells us how big the file is
2. We register the document in the Documents API, which tells us the file parts to upload
3. Our response to ODPC includes the information about the file parts
4. ODPC makes an API call to us for each file part/chunk, which we proxy to the Documents API
5. Once we see that all uploads are completed, we will unlock the document (this is race-condition sensitive)

I've also updated/refactored some existing test helpers since they make life simpler.


src/woo_publications/contrib/documents_api/client.py
```diff
@@ -0,0 +1,138 @@
...
+        description: str = "",
+    ) -> Document:
+        data: EIOCreateBody = {
+            "identificatie": identification,
+            "bronorganisatie": source_organisation,
+            "informatieobjecttype": document_type_url,
+            "creatiedatum": creation_date.isoformat(),
+            "titel": title,
+            "auteur": author,
+            "status": "definitief",
```
bart-maykin, line 71: Can't we make a constant with all the statuses?

src/woo_publications/contrib/documents_api/tests/test_client.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self.assertGreater(len(str(document.uuid)), 0)
+        # we expect a lock to be returend
+        self.assertIsInstance(document.lock, str)
+        self.assertNotEqual(document.lock, "")
+
+        # given the upload size & configuration of Open Zaak, we only expect one part
+        self.assertEqual(len(parts := document.file_parts), 1)
+        self.assertGreater(len(str(parts[0].uuid)), 0)
+        self.assertEqual(parts[0].size, 256_000)
```
bart-maykin, line 71: Didn't know that you could do `256_000`. I assume that this is for readability?

src/woo_publications/contrib/documents_api/client.py
sergei-maertens: There is a type definition for this already and the type checker enforces correctness. See https://github.com/GeneriekPublicatiePlatformWoo/registratie-component/pull/106/files#diff-b2726375446a03f8b580f3224284737684ea2b518e36b2afaf7c0b1ef7cfafc9R7 and https://github.com/GeneriekPublicatiePlatformWoo/registratie-component/pull/106/files#diff-b2726375446a03f8b580f3224284737684ea2b518e36b2afaf7c0b1ef7cfafc9R30

Try adding a typo this value and see what happens :)

src/woo_publications/contrib/documents_api/tests/test_client.py
sergei-maertens: Indeed, since Python 3.9 I believe

src/woo_publications/contrib/documents_api/client.py
bart-maykin: I see that Pyright complains when the typing is incorrect. So in that case looks good to me.

bart-maykin (approved):

PR #109, opened 2024-10-31
:bug: broken uuid unique constraint in  document migration 0004

bart-maykin:
Fixed broken `0004_document_uuid_alter_document_identifier.py` migration by changing migration steps.

- changed initial uuid field addition by not forcing uuids to be unique
- added function to generate unique uuids for document uuid field
- adding alter field migration step to add back unique constraint on document uuid

And added missing uuid field in Document admin. By adding uuid as a read_only field.


src/woo_publications/publications/migrations/0004_document_uuid_alter_document_identifier.py
```diff
@@ -6,6 +6,12 @@
 
 class Migration(migrations.Migration):
 
+    def generate_unique_uuids(apps, _):
```
sergei-maertens, line 9: typically you define these functions outside of the Migration class

sergei-maertens, line 9:
```suggestion
from django.db.migrations.state import StateApps

    def generate_unique_uuids(apps: StateApps, _):
```

You're getting the error/warning here because you defined a method in the class, and the `Migration` class doesn't have a `get_model` method. The type checker thinks that `apps` is your `self` variable, which is why it errors. This combines with the warning that is still left here that your first argument in a method should be named `self`, not `apps`.

The solution is moving it outside of the class and properly type annotate the `apps` parameter :)

src/woo_publications/publications/migrations/0004_document_uuid_alter_document_identifier.py
```diff
@@ -15,6 +21,14 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.AddField(
+            model_name="document",
+            name="uuid",
+            field=models.UUIDField(
+                default=uuid.uuid4, editable=False, verbose_name="UUID"
+            ),
+        ),
+        migrations.RunPython(generate_unique_uuids, migrations.RunPython.noop),  # type: ignore reportArgumentType
```
sergei-maertens, line 30: I think the type checker error here is a real error because you defined a method, not a simple function.

sergei-maertens (changes requested):

src/woo_publications/publications/migrations/0004_document_uuid_alter_document_identifier.py
bart-maykin: The typing isn't necessary when the function is outside of the migration class. So I don't think we need to add it. I am gonna not type apps, unless you disagree with my decision here.

sergei-maertens (approved):

PR #117, opened 2024-11-04
Implement Document create (POST) endpoint

### LINKED ISSUE #19: ODRC: implement Document creation/upload

sergei-maertens:
Fixes #19 (partially)

**Changes**

* Added global configuration
    - [x] Added model + admin
    - [x] Fix audit log mixin breaking initial config record create
    - [x] Add audit/admin smoketests
* [x] implement `POST /api/v1/documenten` endpoint
    * [x] Global config is used to create the document metadata, using `woo_publications.contrib.documenten_api.client`


src/woo_publications/config/models.py
```diff
@@ -0,0 +1,45 @@
...
+        "zgw_consumers.Service",
+        on_delete=models.PROTECT,
+        limit_choices_to={"api_type": APITypes.drc},
+        verbose_name=_("Documents API service"),
+        help_text=_(
+            "The service to use for new document uploads - the metadata and binary "
+            "content will be sent to this API."
+        ),
+        null=True,
+        blank=False,
```
bart-maykin, line 30: What is the reason why these two are different?

src/woo_publications/config/validators.py
```diff
@@ -0,0 +1,39 @@
...
+    Validates that a string value is a valid RSIN number by applying the
+    '11-proef' checking.
+
+    :param value: String object representing a presumably good RSIN number.
+    """
+    # Initial sanity checks.
+    validate_digits(value)
+    if len(value) != RSIN_LENGTH:
+        raise ValidationError(
+            "RSIN moet %s tekens lang zijn." % RSIN_LENGTH, code="invalid-length"
```
bart-maykin, line 27: Any reason why you added multiple dutch error messages which aren't translatable?

sergei-maertens: blind copy-paste from vng-api-common! I'm hopeful that when OAF/commonground-api-common is patched with 3.12 compatibility we can then just import the existing validator

src/woo_publications/config/models.py
sergei-maertens: it's a required field, but django-solo creates the record if it doesn't exist yet, and making it not-nullable crashes with integrity error as there may not be any services defined yet while this record is being created.

`blank=False` ensures that form validation in the admin prevents you from leaving it empty

this pattern is one of the reasons I'm not too fond of django-solo anymore

src/woo_publications/config/validators.py
bart-maykin: It's probably smart to add a TODO comment then so we won't forget about it.

bart-maykin (approved):

PR #122, opened 2024-11-06
Replaced '# type:' to '# pyright:' 

bart-maykin:
Search and replace '# type:' with '# pyright:'


sergei-maertens (approved):
I've updated the PR to make sure the square bracket syntax is used to target specific error classes, otherwise all type errors are ignored which is not what we intend :)

PR #125, opened 2024-11-07
publication status

### LINKED ISSUE #49: ODRC: publicatiestatus
Als gebruiker wil ik kunnen wijzigen welke publicaties en welke documenten wel of niet gepubliceerd moeten zijn/blijven, zodat ik publicaties (deels) kan intrekken waarbij wel later nog gereconstrueerd kan worden wanneer wat online heeft gestaan.

Mogelijk samen oppakken met #64 .

**Acceptatie criteria**

- Op een publicatie kan een `publicatiestatus` worden vastgelegd.
- Op een document kan een  `publicatiestatus` worden vastgelegd.
- De `publicatiestatus` kan een van de drie waarden kennen (enum): `gepubliceerd`, `concept` of `ingetrokken`.
- Het is een verplicht attribuut.
- De default-waarde is `gepubliceerd`.
- De `publicatiestatus` is opgenomen in de API
- Een grafisch datamodel van de database is beschikbaar / bijgewerkt (op GitHub of Read the Docs of elders)

**Taken**

- [x] Add `publicatiestatus` field with specified enum choices to `Publication` model
- [x] Add `publicatiestatus` field with specified enum choices to `Document` model
- [x] Display the fields in the admin changelist page of each model
- [x] Expose the field in the API endpoints/serializers
- [ ] Add explanation of the field in the admin documentation

bart-maykin:
Fixes #49

**Changes**

Added publicatietatus to document and publication.
Added these fields to the admin and api.
Added publication status filters for the document and publication api.

Made sure a publication which is revoked can no longer be updated

Made sure that you cannot create a publication or document which is revoked


src/woo_publications/api/constants.py
```diff
@@ -5,3 +5,9 @@
 class PermissionOptions(models.TextChoices):
     read = "read", _("Read")
     write = "write", _("Write")
+
+
+class PublicationStatusOptions(models.TextChoices):
```
sergei-maertens, line 10: since this is relevant only to the `Publication` and `Document` models, which are both (deliberately :wink: ) defined in the `publications` app, this enum must also be defined in that app. Can you please move it?

src/woo_publications/publications/admin.py
```diff
@@ -6,6 +6,7 @@
 
 from furl import furl
 
+from woo_publications.api.constants import PublicationStatusOptions
```
sergei-maertens, line 9: after moving the enum to this app, this will become a relative import

src/woo_publications/publications/admin.py
```diff
@@ -28,10 +29,11 @@ class DocumentInlineAdmin(admin.StackedInline):
 @admin.register(Publication)
 class PublicationAdmin(AdminAuditLogMixin, admin.ModelAdmin):
     list_display = (
+        "uuid",
```
sergei-maertens, line 32: please put this right before `show_actions` as it was - the documentation now describes that the title is clickable in the admin, this changes that behaviour

The UUID also holds zero meaningful information to humans, it's important for technical reasons and we want to easily be able to copy-paste it (from the list page), but it definitely should not be the first column

src/woo_publications/publications/admin.py
```diff
@@ -49,6 +51,11 @@ class PublicationAdmin(AdminAuditLogMixin, admin.ModelAdmin):
     date_hierarchy = "registratiedatum"
     inlines = (DocumentInlineAdmin,)
 
+    def has_change_permission(self, request, obj=None):
+        if obj and obj.publicatiestatus == PublicationStatusOptions.revoked:
+            return False
+        return True
```
sergei-maertens, line 57: uh this is a security bug, since you need to call the default behaviour instead of just giving permission:

```suggestion
        return super().has_change_permission(self, request, obj=obj)
```

`InformationCategoryAdmin` is also affected by this

the current form completely disregards user permissions in the admin, meaning that any authenticated user can make changes!

The default implementation in django has this:

```py
    def has_change_permission(self, request, obj=None):
        """
        Return True if the given request has permission to change the given
        Django model instance, the default implementation doesn't examine the
        `obj` parameter.

        Can be overridden by the user in subclasses. In such case it should
        return True if the given request has permission to change the `obj`
        model instance. If `obj` is None, this should return True if the given
        request has permission to change *any* object of the given type.
        """
        opts = self.opts
        codename = get_permission_codename("change", opts)
        return request.user.has_perm("%s.%s" % (opts.app_label, codename))
```

src/woo_publications/publications/admin.py
```diff
@@ -28,10 +29,11 @@ class DocumentInlineAdmin(admin.StackedInline):
 @admin.register(Publication)
 class PublicationAdmin(AdminAuditLogMixin, admin.ModelAdmin):
     list_display = (
+        "uuid",
         "officiele_titel",
         "verkorte_titel",
+        "publicatiestatus",
```
sergei-maertens, line 35: please also add it to the `list_filter` :-)

src/woo_publications/publications/api/filters.py
```diff
@@ -3,6 +3,7 @@
 
 from django_filters.rest_framework import FilterSet, filters
 
+from woo_publications.api.constants import PublicationStatusOptions
```
sergei-maertens, line 6: import will become a relative import

src/woo_publications/publications/api/serializers.py
```diff
@@ -72,6 +74,33 @@ class Meta:  # pyright: ignore
...
+            },
+            "publicatiestatus": {
+                "help_text": _(
+                    "\n**Disclaimer**: you can't create a {} document."
+                ).format(PublicationStatusOptions.revoked.label.lower())
+            },
+        }
+
+    def validate(self, attrs):
+        self.instance: Document
```
sergei-maertens, line 89: I'd declare this outside of the method, and perhaps it can be passed as a generic too, I'll experiment with it a little.

either way - `instance` can also be None, I think, but you aren't using it in this method so I don't think it's even necessary in the first place.

```suggestion
```

src/woo_publications/publications/api/serializers.py
```diff
@@ -72,6 +74,33 @@ class Meta:  # pyright: ignore
...
+
+        if self.context["request"].method == "POST":
```
sergei-maertens, line 91: this isn't very elegant, I think better would be:

`is_create = self.instance is None` or if the instance is set: `is_create = self.instance.pk is None`

src/woo_publications/publications/api/serializers.py
```diff
@@ -72,6 +74,33 @@ class Meta:  # pyright: ignore
...
+            if attrs.get("publicatiestatus") == PublicationStatusOptions.revoked:
+                raise serializers.ValidationError(
+                    {
+                        "publicatiestatus": _(
+                            "You cannot create a {} document.".format(
+                                PublicationStatusOptions.revoked.lower()
+                            )
+                        )
+                    }
+                )
```
sergei-maertens, line 101: can we put this in `def validate_publicatiestatus`? We're not doing validation of fields that depend on each other

src/woo_publications/publications/api/serializers.py
```diff
@@ -72,6 +74,33 @@ class Meta:  # pyright: ignore
...
+
+        return super().validate(attrs)
```
sergei-maertens, line 103: usually we make sure to call the super method first:

```py
attrs = super().validate(attrs)
...  # own validation
return attrs
```

src/woo_publications/publications/api/serializers.py
```diff
@@ -155,4 +185,38 @@ class Meta:  # pyright: ignore
...
+            self.instance
+            and self.instance.publicatiestatus == PublicationStatusOptions.revoked
+        ):
+            raise serializers.ValidationError(
+                {
+                    "publicatiestatus": _("You cannot modify a {} publication.").format(
+                        PublicationStatusOptions.revoked.lower()
+                    )
+                }
+            )
```
sergei-maertens, line 220: same remarks as for the publication serializer

src/woo_publications/publications/models.py
```diff
@@ -76,6 +78,12 @@ class Publication(models.Model):
         blank=True,
     )
     omschrijving = models.TextField(_("description"), blank=True)
+    publicatiestatus = models.CharField(
+        _("status"),
+        max_length=12,
+        choices=PublicationStatusOptions.choices,
+        default=PublicationStatusOptions.published,
```
sergei-maertens, line 85: is that a sane default? :thinking: let's discuss with Marco

src/woo_publications/publications/models.py
```diff
@@ -112,6 +120,15 @@ def get_owner(self) -> ActingUser | None:
...
         return log.acting_user[0]
 
+    def clean(self):
+        super().clean()
+        if not self.pk and self.publicatiestatus == PublicationStatusOptions.revoked:
+            raise ValidationError(
+                _("You cannot create a {} publication.").format(
+                    PublicationStatusOptions.revoked.label.lower()
+                )
+            )
```
sergei-maertens, line 130: please respect the order of methods in models to keep things tidy: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#model-style

src/woo_publications/publications/models.py
```diff
@@ -240,6 +263,15 @@ class Meta:  # type: ignore
...
         ]
 
+    def clean(self):
+        super().clean()
+        if not self.pk and self.publicatiestatus == PublicationStatusOptions.revoked:
+            raise ValidationError(
+                _("You cannot create a {} document.").format(
+                    PublicationStatusOptions.revoked.label.lower()
+                )
+            )
```
sergei-maertens, line 273: See above

src/woo_publications/publications/api/serializers.py
```diff
@@ -72,6 +74,33 @@ class Meta:  # pyright: ignore
...
+
+    def validate(self, attrs):
+        self.instance: Document
+
+        if self.context["request"].method == "POST":
+            if attrs.get("publicatiestatus") == PublicationStatusOptions.revoked:
+                raise serializers.ValidationError(
+                    {
+                        "publicatiestatus": _(
+                            "You cannot create a {} document.".format(
```
sergei-maertens, line 96: please always use named arguments, positional arguments are confusing for translators

sergei-maertens (changes requested):

src/woo_publications/publications/models.py
bart-maykin: It definitely is not but it was the described behavior. I think it would indeed be good to discuss this with Marco.

sergei-maertens: discussed and decided to keep this default - changing it later is a minor task so no worries here

sergei-maertens (approved):

PR #126, opened 2024-11-07
Revoke documents when revoking publication

### LINKED ISSUE #64: ODRC: Publicatie intrekken
Als gebruiker wil ik een publicatie in kunnen trekken, zodat ik een foutieve publicatie kan herstellen.

**Toelichting**
In [ODPC #39](https://github.com/GeneriekPublicatiePlatformWoo/Openbaar-Documenten-Publicatie-Component/issues/39) bediscussieerd. Het heeft de voorkeur dat met één API-call een publicatie inclusief documenten ingetrokken kan worden.

Afhankelijk van #49 . Hierin wordt `publicatiestatus` toegevoegd aan de API

**Acceptatie criteria**

- Via de API is het mogelijk om de `publicatiestatus` van een publicatie te wijzigen naar "ingetrokken". (Zie #49)
- Als de `publicatiestatus' van een publicatie wijzigt van "gepubliceerd" of "concept" naar "ingetrokken", dan wordt bij de aanhangige documenten:
  - als de huidige status al "ingetrokken" is, de `publicatiestatus` niet gewijzigd,
  - als de huidige status "gepubliceerd" is, de `publicatiestatus` gewijzigd naar "ingetrokken"
  - als de huidige status "concept" is, de `publicatiestatus` niet gewijzigd,

De consequenties van het wijzigen van de 'publicatiestatus' naar "gepubliceerd" of "concept" vallen niet onder deze story.

**Taken**

make it so when a publication is revoked that the attached documents who are published also get revoked:
- [x] admin
- [x] api 

nice to have:
- [ ] create a unique audit trail event for revoking/publishing documents/publications

bart-maykin:
Fixes #64

**Changes**

- added model function to revoke the published documents of a publication when it gets revoked.
- linked the func to the admin
- linked the func to the api

### DIFF
--- a/src/woo_publications/publications/api/serializers.py
+++ b/src/woo_publications/publications/api/serializers.py
@@ -1,3 +1,4 @@
+from django.db import transaction
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
@@ -212,3 +213,12 @@ class PublicationSerializer(serializers.ModelSerializer[Publication]):
                 )
 
         return value
+
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        publication = super().update(instance, validated_data)
+
+        if validated_data.get("publicatiestatus") == PublicationStatusOptions.revoked:
+            publication.revoke_own_published_documents()
+
+        return publication
--- a/src/woo_publications/publications/api/serializers.py
+++ b/src/woo_publications/publications/api/serializers.py
@@ -4,6 +4,7 @@ from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
 from woo_publications.contrib.documents_api.client import FilePart
+from woo_publications.logging.service import extract_audit_parameters
 from woo_publications.metadata.models import InformationCategory, Organisation
 
 from ..constants import PublicationStatusOptions
@@ -223,6 +224,12 @@ class PublicationSerializer(serializers.ModelSerializer[Publication]):
         publication = super().update(instance, validated_data)
 
         if validated_data.get("publicatiestatus") == PublicationStatusOptions.revoked:
-            publication.revoke_own_published_documents()
+            user_id, user_repr, remarks = extract_audit_parameters(
+                self.context["request"]
+            )
+
+            publication.revoke_own_published_documents(
+                user={"user_id": user_id, "user_repr": user_repr}, remarks=remarks
+            )
 
         return publication
--- a/src/woo_publications/publications/api/serializers.py
+++ b/src/woo_publications/publications/api/serializers.py
@@ -187,10 +187,9 @@ class PublicationSerializer(serializers.ModelSerializer[Publication]):
             "publicatiestatus": {
                 "help_text": _(
                     "\n**Disclaimer**: you can't create a {revoked} publication."
-                    "\n\n**Disclaimer**: when you {revoked} a publication, the attached {published} documents also get's {revoked}."
+                    "\n\n**Disclaimer**: when you revoke a publication, the attached published documents also get revoked."
                 ).format(
                     revoked=PublicationStatusOptions.revoked.label.lower(),
-                    published=PublicationStatusOptions.published.label.lower(),
                 )
             },
         }
@@ -221,6 +220,7 @@ class PublicationSerializer(serializers.ModelSerializer[Publication]):
 
     @transaction.atomic
     def update(self, instance, validated_data):
+        assert instance.publicatiestatus != PublicationStatusOptions.revoked
         publication = super().update(instance, validated_data)
 
         if validated_data.get("publicatiestatus") == PublicationStatusOptions.revoked:
@@ -229,7 +229,7 @@ class PublicationSerializer(serializers.ModelSerializer[Publication]):
             )
 
             publication.revoke_own_published_documents(
-                user={"user_id": user_id, "user_repr": user_repr}, remarks=remarks
+                user={"identifier": user_id, "display_name": user_repr}, remarks=remarks
             )
 
         return publication
--- a/src/woo_publications/publications/api/serializers.py
+++ b/src/woo_publications/publications/api/serializers.py
@@ -186,7 +186,11 @@ class PublicationSerializer(serializers.ModelSerializer[Publication]):
             "publicatiestatus": {
                 "help_text": _(
                     "\n**Disclaimer**: you can't create a {revoked} publication."
-                ).format(revoked=PublicationStatusOptions.revoked.label.lower())
+                    "\n\n**Disclaimer**: when you {revoked} a publication, the attached {published} documents also get's {revoked}."
+                ).format(
+                    revoked=PublicationStatusOptions.revoked.label.lower(),
+                    published=PublicationStatusOptions.published.label.lower(),
+                )
             },
         }
 
--- a/src/woo_publications/publications/models.py
+++ b/src/woo_publications/publications/models.py
@@ -3,6 +3,7 @@ from typing import Callable
 
 from django.core.exceptions import ValidationError
 from django.db import models, transaction
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework.reverse import reverse
@@ -150,6 +151,7 @@ class Publication(models.Model):
             case User():
                 for document in documents:
                     document.publicatiestatus = PublicationStatusOptions.revoked
+                    document.laatst_gewijzigd_datum = timezone.now()
                     audit_admin_update(
                         content_object=document,
                         django_user=user,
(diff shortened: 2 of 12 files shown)


src/woo_publications/publications/api/serializers.py
```diff
@@ -185,7 +187,11 @@ class Meta:  # pyright: ignore
             "publicatiestatus": {
                 "help_text": _(
                     "\n**Disclaimer**: you can't create a {revoked} publication."
-                ).format(revoked=PublicationStatusOptions.revoked.label.lower())
+                    "\n\n**Disclaimer**: when you {revoked} a publication, the attached {published} documents also get's {revoked}."
+                ).format(
+                    revoked=PublicationStatusOptions.revoked.label.lower(),
+                    published=PublicationStatusOptions.published.label.lower(),
+                )
```
sergei-maertens, line 194: This leads to weird sentences in the documentation, keep it simple:

```suggestion
                    "\n\n**Disclaimer**: when you revoke a publication, the attached published documents also get revoked."
                ).format(
                    revoked=PublicationStatusOptions.revoked.label.lower(),
                )
```

src/woo_publications/publications/api/serializers.py
```diff
@@ -212,3 +218,18 @@ def validate_publicatiestatus(
                 )
 
         return value
+
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        publication = super().update(instance, validated_data)
```
sergei-maertens, line 224:
```suggestion
    @transaction.atomic
    def update(self, instance, validated_data):
        assert instance.publicatiestatus != PublicationStatusOptions.revoked
        publication = super().update(instance, validated_data)
```
Our design and assumption now assumes you cannot update a revoked record anymore, we can express this in code with an assertion that serves as documentation which will cause crashes in development if/when behaviour changes (and then is the time to revisit/check our assumptions again).

src/woo_publications/publications/models.py
```diff
@@ -133,6 +138,43 @@ def get_owner(self) -> ActingUser | None:
         assert isinstance(log, TimelineLogProxy)
         return log.acting_user[0]
 
+    def revoke_own_published_documents(
+        self, user: User | ApiUser, remarks: str | None = None
+    ) -> None:
```
sergei-maertens, line 143:
```suggestion
    def revoke(
        self, user: User | ApiUser, remarks: str | None = None
    ) -> None:
```
maybe having a `publication.revoke()` method better encapsulates the meaning of what's happening? you can also add `self.publicatiestatus = PublicationStatus.revoked` to really make this a well encapsulated method, even though for the admin and serializer entrypoints that doesn't make a difference since they already set that field value (so it also doesn't hurt to do it).

src/woo_publications/publications/models.py
```diff
@@ -133,6 +138,43 @@ def get_owner(self) -> ActingUser | None:
...
+        documents = (
+            self.document_set.filter(  # pyright: ignore reportAttributeAccessIssue
```
sergei-maertens, line 145:
```suggestion
            self.document_set.filter(  # pyright: ignore[reportAttributeAccessIssue]
```
the class of errors must go between square brackets

src/woo_publications/publications/models.py
```diff
@@ -133,6 +138,43 @@ def get_owner(self) -> ActingUser | None:
         assert isinstance(log, TimelineLogProxy)
         return log.acting_user[0]
 
+    def revoke_own_published_documents(
+        self, user: User | ApiUser, remarks: str | None = None
+    ) -> None:
+        documents = (
```
sergei-maertens, line 144:
```suggestion
        published_documents = (
```
can't hurt to be more explicit when working with this variable

src/woo_publications/typing.py
```diff
@@ -1,3 +1,7 @@
+from typing import TypedDict
+
 type JSONPrimitive = str | int | float | bool | None
 type JSONValue = JSONPrimitive | JSONObject | list[JSONValue]
 type JSONObject = dict[str, JSONValue]
+
+ApiUser = TypedDict("ApiUser", {"user_id": str, "user_repr": str})
```
sergei-maertens, line 7: why define a new type for what is essentially https://github.com/GeneriekPublicatiePlatformWoo/registratie-component/blob/main/src/woo_publications/logging/typing.py#L4 ? I would just make sure you use `ActingUser` everywhere.

src/woo_publications/publications/models.py
```diff
@@ -133,6 +138,43 @@ def get_owner(self) -> ActingUser | None:
...
+                for document in documents:
+                    document.publicatiestatus = PublicationStatusOptions.revoked
+                    document.laatst_gewijzigd_datum = timezone.now()
+                    audit_api_update(
+                        content_object=document,
+                        user_id=user["user_id"],
+                        user_display=user["user_repr"],
+                        object_data=serialize_instance(document),
+                        remarks=remarks,
+                    )
```
sergei-maertens, line 171: I take it you're doing the `laatst_gewijzigd_datum` thing so that this is properly recorded in the audit trail, since the `auto_now=True` on the field should normally take care of this?

I think I would restructure this a bit:

```py
# get a list of IDs of published documents, make sure to evaluate the queryset so it's not affected by
# the `update` query
document_ids_to_log = list(published_documents.values_list("pk", flat=True))
published_documents.update(publicatiestatus=PublicationStatusOptions.revoked)

# audit log actions
is_django_user = isinstance(user, User)
log_callback = audit_admin_update if is_django_user else audit_api_update
log_extra_kwargs = (
    {"django_user": user} 
    if is_django_user else 
    {"user_id": user["identifier"], "user_display": user["display_name"], "remarks": remarks}
)
for document in Document.objects.filter(pk__in=document_ids_to_log):
    log_callback(
        content_object=document, 
        object_data=serialize_instance(document), 
        **log_extra_kwargs
    )
    
```

src/woo_publications/publications/models.py
```diff
@@ -133,6 +138,43 @@ def get_owner(self) -> ActingUser | None:
...
+
+        # manually update laatst_gewijzigd_datum because bulk_update doesn't trigger save method.
+        Document.objects.bulk_update(
+            documents, ["publicatiestatus", "laatst_gewijzigd_datum"]
+        )
```
sergei-maertens, line 176: if the `auto_now` isn't taken into account automatically, you can still do `published_documents.update(..., laatst_gewijzigd_datum=timezone.now())`

src/woo_publications/publications/tests/test_admin_logging.py
```diff
@@ -175,6 +175,142 @@ def test_admin_update(self):
...
+
+            self.assertEqual(log.extra_data, expected_data)
+
+        form = response.forms["publication_form"]
+        form["publicatiestatus"].select(text=PublicationStatusOptions.revoked.label)
+
+        with freeze_time("2024-09-28T00:14:00-00:00"):
+            response = form.submit(name="_save")
+
+        self.assertEqual(response.status_code, 302)
```
sergei-maertens, line 238: I get a message saying there are validation errors, but I don't see what's wrong/not valid. This test however seems to correctly indicate a succesful revocation :thinking:

src/woo_publications/publications/tests/test_admin_logging.py
```diff
@@ -175,6 +175,142 @@ def test_admin_update(self):
...
+
+        publication.refresh_from_db()
+        published_document.refresh_from_db()
+        concept_document.refresh_from_db()
+        revoked_document.refresh_from_db()
+
+        self.assertEqual(TimelineLogProxy.objects.count(), 3)
+
+        with self.subTest("update publication audit logging"):
+            update_publication_log = TimelineLogProxy.objects.for_object(  # pyright: ignore reportAttributeAccessIssue
```
sergei-maertens, line 248:
```suggestion
            update_publication_log = TimelineLogProxy.objects.for_object(  # pyright: ignore[reportAttributeAccessIssue]
```

sergei-maertens (changes requested):

src/woo_publications/publications/models.py
bart-maykin: I named it `revoke_own_published_documents` specifically because that is exactly what this method does. I don't see the added value of revoking the publication status itself within the function because this will already happen in the admin and api. And it makes it that we can easily implement a feature in the future to revoke all documents attached to a publication if needed.

sergei-maertens: Okay we leave it as is then - but if we ever need to support an atomic action of revoking a publication outside of the admin/API entrypoints (e.g. via a management command or celery task), then we refactor this so that the revocation is a single atomic action captured in one unit of work.

sergei-maertens (approved):

PR #132, opened 2024-11-12
:sparkles: Add sendfile settings

SilviaAmAm:
Needed to serve the certificate from the private media with nginx


src/woo_publications/conf/base.py
```diff
@@ -79,6 +79,10 @@
 # Default (connection timeout, read timeout) for the requests library (in seconds)
 REQUESTS_DEFAULT_TIMEOUT = (10, 30)
 
+SENDFILE_BACKEND = config("SENDFILE_BACKEND", default="django_sendfile.backends.nginx")
+SENDFILE_ROOT = PRIVATE_MEDIA_ROOT
+SENDFILE_URL = PRIVATE_MEDIA_URL
```
SilviaAmAm, line 84: `PRIVATE_MEDIA_ROOT` and `PRIVATE_MEDIA_URL` should be provided by the openapi framework: https://github.com/maykinmedia/open-api-framework/blob/05f58c4520e93dd7364f92891d8712b6617d5233/open_api_framework/conf/base.py#L766

sergei-maertens: I've moved the settings to the correct block in the settings, under the "3RD PARTY LIBRARY SETTINGS" section

sergei-maertens (approved):

PR #161, opened 2024-11-20
:bug: Use 64bit integers for file size metadata

sergei-maertens:
32 bit limits you to roughly 2GB of files, which was not the intent.

The upstream open-zaak/open-zaak#1815 issue is required for this to work properly.


bart-maykin:
@sergei-maertens because this issue is depended on that OZ issue, should I wait with reviewing this issue?

bart-maykin (approved):

PR #170, opened 2024-12-04
custom load organisations command

### LINKED ISSUE #168: Restarting or re-deploying the application resets the active status of organisations
The intent was that this field would not be touched once an organisation is activated and the fixture is loaded again.

bart-maykin:
Fixes #168 

**Changes**

- created management command `load_organisations`
- added tests


sergei-maertens (approved):
Approving, can be merged once the `bin/docker_start.sh` is updated :-)

PR #182, opened 2024-12-13
Tot en met date(time) filters

### LINKED ISSUE #181: Tot en met datum filters

bart-maykin:
Fixes #181

**Changes**

Added lte date(time) filters for documents and publications endpoints.

### DIFF
--- a/src/woo_publications/publications/api/filters.py
+++ b/src/woo_publications/publications/api/filters.py
@@ -56,17 +56,17 @@ class DocumentFilterSet(FilterSet):
         field_name="registratiedatum",
         lookup_expr="lte",
     )
-    creatiedatum_vanaf = filters.DateTimeFilter(
+    creatiedatum_vanaf = filters.DateFilter(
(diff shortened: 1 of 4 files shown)


src/woo_publications/publications/api/filters.py
```diff
@@ -59,6 +66,11 @@ class DocumentFilterSet(FilterSet):
         field_name="creatiedatum",
         lookup_expr="lt",
     )
+    creatiedatum_tot_en_met = filters.DateTimeFilter(
```
sergei-maertens, line 69: the `DateTimeFilter` tripped me here because I had in mind that the database field is just a `DateField`, and after checking, it is!

So the `creatiedatum_vanaf`, `creatiedatum_tot_en_met` and `creatiedatum_tot` filters should be `filters.DateFilter` rather than `DateTimeFilter`.

All the other fields are indeed datetime fields.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #216, opened 2025-03-07
Sending document data to GPP-zoeken

bart-maykin:
Fixes GPP-Woo/GPP-zoeken#12

**Changes**

Added untils function to create the full url for to the download endpoint.
Send the download link from the actual document to the gpp-zoeken document index endpoint with filesize


src/woo_publications/publications/admin.py
```diff
@@ -47,17 +48,25 @@ def sync_to_index(
...
+            document_url = absolute_document_download_uri(
+                request, document_uuid=obj.uuid
+            )
+            transaction.on_commit(
+                partial(
+                    index_document.delay, document_id=obj.pk, download_url=document_url
+                )
+            )
+        else:  # pragma: no cover
+            assert False, "unreachable"
```
bart-maykin, line 69: Opted to use this patern instead of assert_never(model) because of the following error that it caused pyright to return 

` Argument of type "Unknown | type[Publication] | type[Document]" cannot be assigned to parameter "arg" of type "Never" in function "assert_never"
    Type "Unknown | type[Publication] | type[Document]" is not assignable to type "Never"`

src/woo_publications/publications/formset.py
```diff
@@ -4,15 +4,33 @@
...
     def save_new(self, form, commit=True):
         document = super().save_new(form, commit)
-        transaction.on_commit(partial(index_document.delay, document_id=document.pk))
+        if document.publicatiestatus == PublicationStatusOptions.published:
+            transaction.on_commit(
+                partial(index_document.delay, document_id=document.pk)
+            )
         return document
 
     def save_existing(self, form, obj, commit=True):
```
bart-maykin, line 21: We either need to find a way to send the download_url with this or remove the functionality of this. Because if you would update a Document without a download_url the Index will overwrite the `attachemnt.content` to `None`.

sergei-maertens: I've resolved this in 5abc7cd55e1e84dc168281a47b14a6b8bef6112c - it's actually possible with built-in django constructs :) The thread-safety issue is not a concern here because the formset instances are always created newly for each request, it's only the `ModelAdmin` instances themselves that have this issue.

src/woo_publications/publications/utils.py
```diff
@@ -0,0 +1,14 @@
+from uuid import UUID
+
+from django.http import HttpRequest
+from django.urls import reverse_lazy
+
+
+def absolute_document_download_uri(request: HttpRequest, document_uuid: UUID) -> str:
```
sergei-maertens, line 7: I wouldn't make this a utility function, but instead a method on the `Document` model. As far as I can see, this is always being called from places where a `Document` instance is actually in scope. This function is also tightly coupled with the `Document` model, so I'd rather go for:

```py
from rest_framework.reverse import reverse

class Document(...):
    def build_absolute_api_download_uri(self, request: HttpRequest) -> str:
        return reverse("api:document_download", kwargs={"uuid": self.uuid}, request=request)
```

The DRF `reverse` version already handles making URLs fully qualified and it also supports any potential version negotiation from API endpoints (if we'll ever need that in the future).

sergei-maertens (changes requested):
Nice work, some minor polishing to be done and then it's ready to merge.

PR #221, opened 2025-03-26
Onderwerpen model + admin

bart-maykin:
Partially fixes #205

**Changes**

- Added the model Subject with all base fields besides the connection to publications
- Added the admin page without the publication connection and functionality to send data to the search component
- wrote basic admin and admin audit log tests and changed some duplicate admin filter test for the Publication and Document admin.
- started on writing the documentation for the newly added admin page. 

TODO (can be done in a different pr):
- [x] add missing documentation for the field `promoot` -> #227
- [ ] add translations after all the help_texts and such are approved.
- [x] update the changelog -> #228

PR #222, opened 2025-03-26
document search field tests improvement

bart-maykin:
Dependend on #221 

**Changes**

Improved the search field test for the document admin


sergei-maertens (commented):
Will be okay to merge once the previous PR has gone through.

PR #228, opened 2025-03-28
:memo: added topics to the changelog

bart-maykin:
Updated the changelog for the Topics addition


sergei-maertens:
Discussed on Slack - we manage changelogs at release time :)

PR #233, opened 2025-04-01
Stream large file part uploads

### LINKED ISSUE #232: Ensure large file uploads are streamed to the Documenten API
Large file uploads are now read into memory to perform the multipart upload, which is a problem if the container/node has memory constraints (5GB container memory requests are... excessive). It's definitely a problem if/when there are concurrent large file uploads (which will happen when uploading chunks in parallel).

Out of the box, requests does not support streaming file uploads that also include additional form-data. There seem to be options though:

* using `MultipartEncoder` from requests-toolbelt (preferred)
* check if `httpx` has support and switch library for these uploads

I'm classifying this as a bug since it impacts the use of large file uploads, and so we can include it in patch releases.

sergei-maertens:
Fixes #232 

**Changes**

* Leveraged multipart encoder from requests-toolkit to stream large files so we don't have to load them into memory.

The chunks are read 16MB at a time, via urllib3's default configuration. I haven't checked if we can override this, but I assume the
defaults are okay for now.

### DIFF
--- a/src/woo_publications/utils/multipart_encoder.py
+++ b/src/woo_publications/utils/multipart_encoder.py
@@ -1,0 +1,501 @@
+# SPDX-FileCopyrightText: Copyright 2014 Ian Cordasco, Cory Benfield
+# SPDX-License-Identifier:  Apache-2.0
+#
+# The code in this module was taken from requests-toolbelt:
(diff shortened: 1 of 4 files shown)


src/woo_publications/utils/multipart_encoder.py
sergei-maertens, line 1: TODO: I need to exclude this file from coverage reports since it's vendored from another package

bart-maykin (approved):

PR #247, opened 2025-04-11
apply retention policy to publication creation

### LINKED ISSUE #200: Apply retention policy to new publications
**! Let op, #198 en #199 zijn randvoorwaardelijk!**

_Als informatiebeheerder wil ik dat de bewaartermijn automatisch wordt toegekend aan iedere nieuwe publicatie, zodat we t.z.t. de vernietiging of overbrenging kunnen uitvoeren._

**Design keuzes**
- We gaan ervan uit dat de bewaartermijn in alle gevallen gaat lopen vanaf de publicatiedatum. Aangezien we de publicatiedatum en de registratiedatum niet van elkaar hebben gescheiden, gaan we uit van de registratiedatum.

**Acceptatie criteria**

- Onderstaande criteria gelden zowel voor de creatie van publicaties via de admin-interface als via de API.
- Bij de creatie van een _nieuwe_ publicatie:
  - Wordt de waarde `Bron bewaartermijn` gekopieerd van de gekoppelde `informatiecategorie`
  - Wordt de `selectiecategorie` gekopieerd van de gekoppelde `informatiecategorie`
  - Wordt de `archiefnominatie` gekopieerd van de gekoppelde `informatiecategorie`
  - Wordt de `archiefactiedatum` berekend door de `registratiedatum` op te tellen met het aantal jaren in `bewaartermijn (in jaren)`.
  - Wordt de `toelichting bewaartermijn` gekopieerd van de gekoppelde `informatiecategorie`.
- Als er meer dan één informatiecategorieën gekoppeld zijn:
  - (stap 1) dan hebben bewaartermijnen waarbij `archiefnominatie` = "blijvend_bewaren"  altijd voorrang op de bewaartermijnen waarbij `archiefnominatie` = "vernietigen",
  - (stap 2a) als er een of meerdere bewaartermijnen zijn waarbij `archiefnominatie` = "blijvend_bewaren" dan geldt de kortste `bewaartermijn (in jaren)`
  - (stap 2b) als er alleen nog bewaartermijnen zijn waarbij `archiefnominatie` = "vernietigen" dan geldt de langste `bewaartermijn (in jaren)`.
- Documentatie is bijgewerkt (taak voor PO)

**Taken**

- [ ]

bart-maykin:
Fixes #200

**Changes**

- Added function to sync up the retention fields from the IC to the publication in the admin and api.
- Added a helper function to determine which IC from a list is applicable to use for filling in the retention fields.


src/woo_publications/publications/admin.py
```diff
@@ -223,6 +225,20 @@ def save_model(
         is_published = new_status == PublicationStatusOptions.published
         is_revoked = new_status == PublicationStatusOptions.revoked
 
+        if not change:
+            retention_ic = get_retention_informatie_category(
+                form.cleaned_data["informatie_categorieen"]
+            )
+
+            if retention_ic:
```
bart-maykin, line 233: Could change to an assert because this should always have a record but this seemed saver?

src/woo_publications/publications/api/serializers.py
```diff
@@ -344,6 +377,26 @@ def update(self, instance, validated_data):
...
         return publication
 
+    @transaction.atomic
+    def create(self, validated_data):
+        ic_queryset = InformationCategory.objects.filter(
+            id__in=[ic.pk for ic in validated_data["informatie_categorieen"]]
+        )
+        retention_ic = get_retention_informatie_category(ic_queryset)
+
+        if retention_ic:
```
bart-maykin, line 387: Could change to an assert because this should always have a record but this seemed saver?

src/woo_publications/publications/admin.py
sergei-maertens, line 233: I think it's better to tap into the `save_related` method - you still have the `change` argument to see if it was a create or change, but you have the benefit that you can query `publication.informatie_categorieen.all()` and that will have been saved in the database already, so you don't need to mess around with `form.cleaned_data`

I think I'd then indeed opt for an `assert ic is not None`, with a comment saying that the information categories for a publication require at least one entry according to the model definition so we know where the assumption is coming from.

src/woo_publications/publications/admin.py
```diff
@@ -223,6 +225,20 @@ def save_model(
...
+                obj.bron_bewaartermijn = retention_ic.bron_bewaartermijn
+                obj.selectiecategorie = retention_ic.selectiecategorie
+                obj.archiefnominatie = retention_ic.archiefnominatie
+                obj.archiefactiedatum = date.today() + relativedelta(
+                    years=retention_ic.bewaartermijn
+                )
+                obj.toelichting_bewaartermijn = retention_ic.toelichting_bewaartermijn
```
sergei-maertens, line 240: I think it'd be much cleaner to implement a model method to encapsulate the logic and calculation:

```py
from django.utils.timezone import localdate

def apply_retention_policy(self) -> None:
    """
    Calculate and save the retention policy for the publication.
    """
    information_category = get_retention_informatie_category(self.information_categorieen.all())
    assert information_category is not None, "A publication must have at least one information category"
    self.bron_bewaartermijn = information_category.bron_bewaartermijn
    self.selectiecategorie = information_category.selectiecategorie
    self.archiefnominatie = information_category.archiefnominatie
    # make sure to calculate and store the Europe/Amsterdam date (!)
    self.archiefactiedatum = localdate(self.registratiedatum + relativedelta(years=retention_ic.bewaartermijn))
    self.save()
```

and then call this method in the admin and the API endpoint(s)

The benefit is that later you'll be able to call this method again if there's a change in information categories and it's much simpler without needing to repeat yourself, + you can properly unit-test it at the model layer.

src/woo_publications/publications/api/serializers.py
```diff
@@ -302,6 +306,35 @@ class Meta:  # pyright: ignore
                     revoked=PublicationStatusOptions.revoked.label.lower(),
                 )
             },
+            "bron_bewaartermijn": {
+                "help_text": _(
+                    "The source of the retention policy (example: Selectielijst gemeenten 2020)."
+                    "\n\n**Disclaimer** during creation the input will be overwritten by the linked information category."
```
sergei-maertens, line 312:
```suggestion
                    "\n\n**Note** on create, manually provided values are ignored and overwritten by the "
                    "automatically derived parameters from the related information categories."
```

src/woo_publications/publications/api/serializers.py
```diff
@@ -344,6 +377,26 @@ def update(self, instance, validated_data):
...
+            validated_data["selectiecategorie"] = retention_ic.selectiecategorie
+            validated_data["archiefnominatie"] = retention_ic.archiefnominatie
+            validated_data["archiefactiedatum"] = date.today() + relativedelta(
+                years=retention_ic.bewaartermijn
+            )
+            validated_data["toelichting_bewaartermijn"] = (
+                retention_ic.toelichting_bewaartermijn
+            )
+
+        return super().create(validated_data)
```
sergei-maertens, line 398: It'd be much simpler using the suggested method above:

```py
@transaction.atomic
def create(self, validated_data):
    publication = super().create(validated_data)
    publication.apply_retention_policy()
    return publication
```

src/woo_publications/publications/tests/test_publications_admin.py
```diff
@@ -244,15 +251,173 @@ def test_publications_admin_create(self):
...
                 added_item.archiefnominatie, ArchiveNominationChoices.retain
             )
-            self.assertEqual(str(added_item.archiefactiedatum), "2025-01-01")
-            self.assertEqual(added_item.toelichting_bewaartermijn, "extra data")
+            self.assertEqual(str(added_item.archiefactiedatum), "2029-09-25")
+            self.assertEqual(added_item.toelichting_bewaartermijn, "toelichting")
+
+    def test_publication_admin_create_retention_field_with_both_archive_nomination_choices(
+        self,
+    ):
```
sergei-maertens, line 265: These tests would be much better suited as model unit tests with the suggested method above - you have less setup to do and the assertions become simpler.

src/woo_publications/publications/tests/test_utils.py
sergei-maertens, line 1: I think this helper method is an implementation detail and writing the model method tests should cover all these situations. For the rest, the type checker can prevent silly mistakes and that will allow us to refactor in the future without also having to update tests.

sergei-maertens (changes requested):

src/woo_publications/publications/admin.py
bart-maykin: This was my initial route as well, but then we needed to have 2 queries to save all the data. So I thought it was better to stay away from it.

sergei-maertens: it's fine, this is not a performance critical code path.

src/woo_publications/publications/tests/test_publications_admin.py
bart-maykin: I kept these old test because we need to have some test proving that it is applied in the admin, but I also added model tests for this.

src/woo_publications/publications/admin.py
```diff
@@ -303,6 +303,17 @@ def delete_queryset(
...
+    def save_related(
+        self,
+        request: HttpRequest,
+        form: forms.Form,
+        formsets: forms.BaseModelFormSet,
+        change: bool,
+    ):
+        super().save_related(request, form, formsets, change)
+        if not change:
+            form.instance.apply_retention_policy()  # pyright: ignore[reportAttributeAccessIssue]
```
sergei-maertens, line 315: Better is:

```suggestion
        if not change:
            assert isinstance(form.instance, Publication)
            form.instance.apply_retention_policy()
```

src/woo_publications/publications/archiving.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_retention_informatie_category(
+    informatie_categorieen: QuerySet[InformationCategory],
+) -> InformationCategory | None:
+    """
+    Determine the most appropriate information category for archiving parameter derivation.
+
+    The following logic is applied:
+    - Information categories with :arg:`archiefnominatie` ``retain`` will always be used if present.
```
sergei-maertens, line 14:
```suggestion
    The following logic is applied:

    - Information categories with :arg:`archiefnominatie` ``retain`` will always be used if present.
```
the blank line is required for proper RST rendering I think.

src/woo_publications/publications/tests/test_publication_model.py
```diff
@@ -0,0 +1,109 @@
...
+            publication = PublicationFactory.create(
+                informatie_categorieen=[ic1, ic2, ic3, ic4, ic5]
+            )
+
+        # sanity check that the fields were empty
+        self.assertEqual(publication.bron_bewaartermijn, "")
+        self.assertEqual(publication.selectiecategorie, "")
+        self.assertEqual(publication.archiefnominatie, "")
+        self.assertIsNone(publication.archiefactiedatum)
+        self.assertEqual(publication.toelichting_bewaartermijn, "")
```
sergei-maertens, line 60: I prefer plain `asserts` for sanity checks, as now it's visually not immediately clear what the test assertions are.

src/woo_publications/publications/tests/test_publication_model.py
```diff
@@ -0,0 +1,109 @@
...
+
+        publication.apply_retention_policy()
+
+        publication.refresh_from_db()
+        self.assertEqual(publication.bron_bewaartermijn, "bewaartermijn 1")
+        self.assertEqual(publication.selectiecategorie, "1.0.1")
+        self.assertEqual(publication.archiefnominatie, ArchiveNominationChoices.retain)
+        self.assertEqual(
+            str(publication.archiefactiedatum), "2029-09-24"
+        )  # 2024-09-24 + 5 years
```
sergei-maertens, line 70: why not just `self.assertEqual(publication.archiefactiedatum, date(2029, 9, 24))`? This indirection via a string is not necessary.

src/woo_publications/publications/tests/test_archiving.py
```diff
@@ -0,0 +1,74 @@
+from django.test import TestCase
+
+from woo_publications.constants import ArchiveNominationChoices
+from woo_publications.metadata.models import InformationCategory
+from woo_publications.metadata.tests.factories import InformationCategoryFactory
+
+from ..archiving import get_retention_informatie_category
+
+
+class GetRetentionInformationCategoryTest(TestCase):
```
sergei-maertens, line 10: I would rather see these cases only in the model tests and not any unit test for the `get_retention_informatie_category` helper. It is an implementation detail - as long as it's 100% covered via the model tests, we're good.

Now, if we change this function, move it, or delete it, we have to update these tests, even though functionally nothing changed. (Unit) tests must exist for the "public API", like the model methods, but these helpers that are only used to support the public API are not part of the contract to test and they're not intended to used anywhere else.

src/woo_publications/publications/tests/test_publications_admin.py
sergei-maertens: One minimal "create" test that proves admin actions calculate the archiving details are sufficient. On top of that, you want one "update" test where the archiving parameters are overridden, and assert that the model method is not invoked again. All the rest can go. And with minimal I also mean: just a single IC is enough.

All these extra tests have no added value over the model tests, but they will cause problems when behaviour changes and we have to update them. It's the testing pyramid:

1. Have unit tests (the model method) for the various edge cases
2. Have an integration tests proving that the model method is invoked when it should (the admin test)

src/woo_publications/publications/admin.py
bart-maykin: I did try that but doesn't fix the error `Cannot access attribute "instance" for class "Form"
    Attribute "instance" is unknown (reportAttributeAccessIssue)`

src/woo_publications/publications/tests/test_publications_admin.py
bart-maykin: I won't write the test for the update, because this will be done in #201. Especially because the logic for when it does and doesn't need to be invoked will change.

sergei-maertens (approved):

PR #250, opened 2025-04-15
Apply retention policy when updating publication

### LINKED ISSUE #201: Update retention policy when updating a publication
**! Let op! #200 is randvoorwaardelijk!**

_Als informatiebeheerder wil ik dat het wijzigen van de informatiecategorieën op een publicatie leidt tot een herziening van de bewaartermijn, zodat deze altijd klopt._

**Acceptatie criteria**

- Onderstaande criteria gelden zowel voor de mutatie van publicaties via de admin-interface als via de API.
- Als bij een publicatie een informatiecategorie wordt toegevoegd en/of verwijderd, dan worden de bewaartermijn opnieuw bepaald en ingevuld conform de procedure zoals beschreven in #200 .
- Documentatie is bijgewerkt.

**Acties**

- [ ]

bart-maykin:
Fixes #201

**Changes**

Applied the retention field sync logic from #200 during the update of linked IC fields in the admin and api.


src/woo_publications/publications/admin.py
```diff
@@ -311,7 +311,7 @@ def save_related(
         change: bool,
     ):
         super().save_related(request, form, formsets, change)
-        if not change:
+        if not change or change and "informatie_categorieen" in form.changed_data:
```
sergei-maertens, line 314:
```suggestion
        if not change or (change and "informatie_categorieen" in form.changed_data):
```
for readability, can you make the precedence a bit more clear?

src/woo_publications/publications/api/serializers.py
```diff
@@ -376,6 +376,9 @@ def update(self, instance, validated_data):
                 user={"identifier": user_id, "display_name": user_repr}, remarks=remarks
             )
 
+        if validated_data.get("informatie_categorieen"):
+            publication.apply_retention_policy()
```
sergei-maertens, line 380: aren't these always going to be provided in PUT calls? so that would negate manual overwrites from GPP-app

src/woo_publications/publications/tests/test_publications_api.py
```diff
@@ -1510,40 +1517,68 @@ def test_partial_update_publication(self):
             "api:publication-detail",
             kwargs={"uuid": str(publication.uuid)},
         )
-        data = {
-            "officieleTitel": "changed offical title",
-        }
+        with self.subTest("changing single field only updates dat field"):
```
sergei-maertens, line 1520:
```suggestion
        with self.subTest("changing single field only updates that field"):
```

src/woo_publications/publications/api/serializers.py
bart-maykin: The informatie_categorieen is atm a required field so it will always be provided in the PUT call. But when you are using a patch call this isn't the case, so we don't want to accidentally trigger it in those cases.

sergei-maertens: yeah but I don't think gpp-app uses PATCH calls. Functionally, if I make a PUT call where I change just the description of a publication, this should not re-trigger retention policy calculation. It's entirely valid to use a PUT call where you change the archiefactiedatum manually, no? Even though the UI doesn't enable this right now, I think this behaviour needs to be covered in the API now because changing this later will be a breaking change and require a new major version.

sergei-maertens (approved):

PR #251, opened 2025-04-16
Feature/207 topic image field

### LINKED ISSUE #207: Add graphic image to topics
**Let op! [GPP-publicatiebank #206](https://github.com/GPP-Woo/GPP-publicatiebank/issues/206) en [GPP-publicatiebank #211](https://github.com/GPP-Woo/GPP-publicatiebank/issues/211) zijn randvoorwaardelijk!**

**Acceptatie criteria**

- Vanuit de admin-interface wil ik bij een _onderwerp_ één afbeelding / foto kunnen uploaden.
- Het veld hiervoor is verplicht
- De afbeelding kan vervangen worden.
- Via de API kan de afbeelding bij een _onderwerp_ geraadpleegd worden, zodat deze getoond kan worden, bijv. in het GPP-burgerportaal.
- Graag in overleg met ontwikkelaars GPP-burgerportaal bepalen:
  - ondersteuning van bestandstypes (jpeg, gif, etc.)
  - max. bestandsgrootte
- Logging / audittrail is bijgewerkt
- Documentatie is bijgewerkt.

**Taken**

- [ ] Add image field to model
- [ ] Expose image field in API endpoints
- [ ] Ensure that Helm charts use the `settings.MEDIA_ROOT` etc. and has volumes mounted to store the images

bart-maykin:
Fixes #207

**Changes**

Added the image field to the topics.
- Added 3 validators to check image size, width/height and extension
- Added the image field to the admin, where images can be uploaded
- Added image field to the topic's api endpoint as a readonly field which returns the image path


src/woo_publications/logging/serializing.py
```diff
@@ -17,6 +18,8 @@ def model_to_dict(instance):
         match f:
             case models.ManyToManyField():
                 value = [obj.pk for obj in value]
+            case models.ImageField():
+                value = base64.b64encode(value.file.read()).decode("ascii")
```
bart-maykin, line 22: I don't know if this is a good way to solve this, but I thought this with maybe a new template to display the image properly would do the trick? If you agree then I will look into the template.

src/woo_publications/utils/tests/test_validators.py
```diff
@@ -0,0 +1,69 @@
...
+
+from woo_publications.publications.tests.factories import TopicFactory
+
+from ..validators import (
+    image_extension_validator,
+    max_file_size_validator,
+    max_file_width_and_height_validator,
+)
+
+
```
bart-maykin, line 13: I don't particularly like these tests but because the validator's need the value to be an `ImageFieldFile` this seems to be the easiest way to do this.

src/woo_publications/conf/base.py
```diff
@@ -80,6 +81,25 @@
...
+        "webp",
+    ],
+)
+assert set(ALLOWED_IMG_EXTENSIONS) <= set(
+    validators.get_available_image_extensions()
+), "img file type not supported"
+
+MAX_FILE_SIZE = config("MAX_FILE_SIZE", default=1_000_000)
+MAX_FILE_HEIGHT = config("MAX_FILE_HEIGHT", default=300)
+MAX_FILE_WIDTH = config("MAX_FILE_WIDTH", default=300)
```
bart-maykin, line 101: Probably needs some extra context, but I don't know if we want to go this route or not.

src/woo_publications/utils/tests/test_validators.py
sergei-maertens: they actually don't, they must implement the `ImageFile` interface, which `ImageFieldField` subclasses :)

So, updating your type annotations to use `django.core.files.images.ImageFile` instead should be sufficient, and then You can generate these in memory quite easily:

```py
from io import BytesIO
from django.core.files.images import ImageFile
from PIL import Image

image_file = ImageFile(BytesIO(), name="example.png")
Image.new("RGBA", (16, 16), (255, 0, 0, 255)).save(image_file, format="PNG")
>>> image_file.width
16
>>> image_file.height
16
```

sergei-maertens (changes requested):

src/woo_publications/conf/base.py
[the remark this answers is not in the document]
sergei-maertens: let's set the defaults to 600px like discused on Slack and then we're good to go!

src/woo_publications/conf/base.py
```diff
@@ -80,6 +81,38 @@
...
+    "ALLOWED_IMG_EXTENSIONS",
+    default=[
+        "jpg",
+        "jpeg",
+        "png",
+        "gif",
+        "webp",
+    ],
+    help_text=_("The allowed image extensions that we support."),
+)
```
sergei-maertens, line 95: iirc there's a `group` or `section` parameter as well to group these together. A heading like "Image upload settings" would be good.

src/woo_publications/publications/tests/test_topic_api.py
```diff
@@ -42,18 +44,19 @@ def test_403_when_audit_headers_are_missing(self):
             self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
     def test_api_key_result_in_401_with_wrong_credentials(self):
-        publication = PublicationFactory.create()
+        topic = TopicFactory.create()
```
sergei-maertens, line 47: this testcase will also need a `MEDIA_ROOT` override, or make the factory not generate an image in this instance :)

src/woo_publications/utils/tests/test_validators.py
```diff
@@ -0,0 +1,81 @@
...
+            image_file.file, format="PNG"
+        )
+
+        with self.assertRaisesMessage(
+            ValidationError,
+            _("File size exceeds max size of {max_img_size}.").format(
+                max_img_size=filesizeformat(10)
+            ),
+        ):
+            max_img_size_validator(image_file)
```
sergei-maertens, line 38: So - you did everything right IMO. This breaks because Django's `assertRaisesMessage` is a dirty hack. It checks the error on the string representation of the exception. The string representation of the exception contains the string representation of the error message, and that one escapes the backslash. That's how you end up with the double backslash.

The suggestion below avoids this by using the known interface of `ValidationError` and we compare the exact error message instead of relying on the exception string representation.

```suggestion
    @override_settings(MAX_IMG_SIZE=10)
    def test_size_larger_then_max(self):
        image_file = ImageFile(BytesIO(), name="example.png")
        assert isinstance(image_file.file, BytesIO)
        Image.new("RGBA", (16, 16), (255, 0, 0, 255)).save(
            image_file.file, format="PNG"
        )

        expected_error = _("File size exceeds max size of {max_img_size}.").format(
            max_img_size=filesizeformat(10)
        )

        with self.assertRaises(ValidationError) as exc_cm:
            max_img_size_validator(image_file)
        self.assertEqual(exc_cm.exception.message, expected_error)
```
this works around the problem :)

sergei-maertens (approved):

Issue #267, opened 2025-04-30
Ensure that the KOOP/TOOI idenfifiers are used for documenthandelingen


sergei-maertens:
The absence of these correct identifiers now results in the sitemap XML not passing the XSD schema validation, breaking the harvester.

PR #269, opened 2025-04-30
:bug: [#267] Add missing waardelijst identifiers for documenthandelingen

sergei-maertens:
Fixes #

**Changes**

* Added constant mapping enum to TOOI identifier
* Include identifier in serializer/API output for sitemap generation

### DIFF
--- a/src/woo_publications/publications/constants.py
+++ b/src/woo_publications/publications/constants.py
@@ -1,3 +1,5 @@
+from typing import Mapping
+
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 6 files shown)


src/woo_publications/publications/constants.py
```diff
@@ -12,3 +14,18 @@ class DocumentActionTypeOptions(models.TextChoices):
...
+
+
+# Taken from PLOOI documenthandelingen on
+# https://standaarden.overheid.nl/tooi/waardelijsten/expression?lijst_uri=https%3A%2F%2Fidentifier.overheid.nl%2Ftooi%2Fset%2Fccw_plooi_documenthandelingen%2F2
+# and https://repository.officiele-overheidspublicaties.nl/waardelijsten/ccw_plooi_documenthandelingen/2/json/ccw_plooi_documenthandelingen_2.json
+DOCUMENT_ACTION_TOOI_IDENTIFIERS: Mapping[DocumentActionTypeOptions, str] = {
+    DocumentActionTypeOptions.signed: "https://identifier.overheid.nl/tooi/def/thes/kern/c_e1ec050e",
+    DocumentActionTypeOptions.received: "https://identifier.overheid.nl/tooi/def/thes/kern/c_dfcee535",
+    DocumentActionTypeOptions.declared: "https://identifier.overheid.nl/tooi/def/thes/kern/c_641ecd76",
+}
```
bart-maykin, line 26: These values can technically change. Do we want to do something with that? Now or in the future?

sergei-maertens: when they change, we will update accordingly. these value lists are versioned and it's the same process if a new version of the organisations/information categories value lists is published that we need to load :)

bart-maykin (approved):

PR #284, opened 2025-05-09
Django 5.2

### LINKED ISSUE #277: Upgrade to Django 5.2 LTS
Last month, Django version 5.2 was released, which is a Long Term Support release. Currently we use Django 4.2, the previous LTS release, which will be maintained until April 2026.

To keep receiving security fixes in the future and ability to use new features, we must consider/plan the upgrade to Django 5.2 - it is regular maintenance.

bart-maykin:
Fixes #277 

**Changes**

Updated Django to 5.2

### DIFF
--- a/src/woo_publications/conf/ci.py
+++ b/src/woo_publications/conf/ci.py
@@ -26,7 +26,7 @@ CACHES.update(
 mute_logging(LOGGING)
 
 # don't spend time on password hashing in tests/user factories
-PASSWORD_HASHERS = ["django.contrib.auth.hashers.UnsaltedMD5PasswordHasher"]
+PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
 
 ENVIRONMENT = "CI"
 
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -48,7 +48,6 @@ PostgreSQL is the database used. PostgreSQL 13 and newer are supported.
    =============  ==========================
    PostgreSQL     Status
    =============  ==========================
-   13             Supported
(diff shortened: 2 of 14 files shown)


bart-maykin:
@sergei-maertens Because of the change in the admin for url fields (make the urls now a clickable link in the admin page), do we want to ensure that the identifier for the meta models behave differently depending on if we add them or we load them in? Because we create fake urls for them, when the ones that get loaded in get redirected to the waardenlijst.

sergei-maertens:
> @sergei-maertens Because of the change in the admin for [url fields](https://docs.djangoproject.com/en/5.2/releases/5.2/#django-contrib-admin), do we want to ensure that the identifier for the meta models behave differently depending on if we add them or we load them in? Because we create fake urls for them, when the ones that get loaded in get redirected to the waardenlijst.

No problem to leave it as is. If we get feedback, we can spend some effort.

src/woo_publications/conf/ci.py
```diff
@@ -25,9 +25,6 @@
 # shut up logging
 mute_logging(LOGGING)
 
-# don't spend time on password hashing in tests/user factories
-PASSWORD_HASHERS = ["django.contrib.auth.hashers.UnsaltedMD5PasswordHasher"]
```
sergei-maertens, line 29: instead make it:

```py
PASSWORD_HASHERS = ["django.contrib.auth.hashers.MD5PasswordHasher"]
```
this one *does* still exist and is faster than all the other password hashes, which is the reason why this is specified in the CI settings: faster tests

sergei-maertens (approved):

PR #305, opened 2025-06-06
Feature/275 extra descriptions

### LINKED ISSUE #275: Add `Omschrijving` to information categories
**Acceptatie criteria**
- Aan informatiecategorieën wordt een het attribuut `Omschrijving` toegevoegd.
  - Deze is optioneel en kan in de admin-interface ingevuld/gewijzigd worden.
- De `Omschrijving` kan via de API geraadpleegd worden (GET)
- Documentatie is bijgewerkt.

**Acties**

- [ ]

bart-maykin:
Fixes #275

**Changes**

- Added optional field `omschrijving` for the information_categories.
- Added it in the admin and api
- updated documentation


sergei-maertens (commented):
How does this field interact with loading the information categories from the waardenlijsten?

bart-maykin:
> How does this field interact with loading the information categories from the waardenlijsten?

Not well, seems that it completely overwrites the data. Same goes for the retention fields on the same model and the is_active boolean on the organisations.

We should create a separate issue for this.

sergei-maertens:
I've checked it, organisations is fine because we have a custom management command: https://github.com/GPP-Woo/GPP-publicatiebank/blob/main/src/woo_publications/metadata/management/commands/load_organisations.py

> We should create a separate issue for this.

This is not a good option - the fixtures are loaded on every deploy, so every time I deploy the publicatiebank, all the archiving parameters are currently reset and this description would we wiped out. I will do it however because otherwise this PR stays open too long.

PR #306, opened 2025-06-06
disable required validation for concept publications

### LINKED ISSUE #264: Create concept-document with incomplete data
**_Let op! #266 is randvoorwaardelijk!_**

**Acceptatie criteria**

- Wanneer een nieuw document toegevoegd wordt aan een concept-publicatie (zie #263):
  - Dan krijgt deze ook de `status` "concept" (zie #266).
  - Dan wordt de registratie ook geaccepteerd en opgeslagen wanneer **_niet_** alle verplichte velden zijn ingevuld.
    - Wèl verplicht blijft de gekoppelde `publicatie`.
    - Wèl verplicht blijven de header parameters in de API-call.
    - Als een veld wel een waarde krijgt, dan moet deze wel in het juiste format zijn (bijv. string, enumeratie of datum)
- Een bestaand concept-document kan gewijzigd worden met behoud van bovenstaande criteria.
- Al het bovenstaande werkt zowel via de API als via de admin-interface.
- Documentatie is bijgewerkt

**Acties*

- [ ]

bart-maykin:
Fixes #264 

**Changes**

- updated publication model to let make all fields but officieleTitel optional
- Made sure that the publication admin/api sets all fields but officieleTitel as not required during concept creation/update
- Updated documentation


src/woo_publications/publications/tests/test_publications_api.py
```diff
@@ -1342,12 +1342,12 @@ def test_create_publication(self):
                 "verkorteTitel": "one",
                 "omschrijving": "Lorem ipsum dolor sit amet, "
                 "consectetur adipiscing elit.",
-                # These values will be overwritten
-                "bronBewaartermijn": "THIS VALUE WON'T BE USED",
-                "selectiecategorie": "THIS VALUE WON'T BE USED",
+                # These values will not be overwritten because the status is concept
+                "bronBewaartermijn": "THIS VALUE WILL BE USED",
+                "selectiecategorie": "THIS VALUE WILL BE USED",
```
CharString, line 1347: I don't have the domain knowledge to asses this.

I expected this to already be used, because I thought you either don't specify these fields and the will be applied from the informatieCategorieen defaults, or you do specify them, and they get used.

src/woo_publications/publications/tests/test_publications_api.py
```diff
@@ -1891,3 +1891,120 @@ def test_updated_publication_schedules_index_task(
...
+            "officieleTitel": "",
+        }
+
+        response = self.client.put(url, data, headers=AUDIT_HEADERS)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(
+            response.json(),
+            {
+                "officieleTitel": [_("This field cannot be null.")],
```
CharString, line 1951: `""` becomes `null` somewhere? Why?

src/woo_publications/publications/api/serializers.py
```diff
@@ -641,6 +658,8 @@ def create(self, validated_data):
                     f"Unexpected creation publicatiestatus: {publicatiestatus}"
                 )
 
+        publication.save()
+
         publication.apply_retention_policy()
```
CharString, line 663: Extra `save()`? Maybe remove the save from the apply and save afterwards? Or make the apply function accept a `save: bool=True`. If that makes more sense from the other call sites.

CharString (changes requested):

src/woo_publications/publications/models.py
```diff
@@ -170,6 +170,8 @@ class Publication(ConcurrentTransitionMixin, models.Model):
         help_text=_("The organisation which publishes the publication."),
         limit_choices_to={"is_actief": True},
         on_delete=models.CASCADE,
+        # only allow null for concept
+        null=True,
```
sergei-maertens, line 174: you can express this in a database constraint!

```py
class Meta:
    constraints = [
        models.CheckConstraint(
            condition=~models.Q(publicatiestatus=PublicationStatusOptions.concept) & models.Q(publisher__isnull=False),
            name="non_concept_publication_required_publisher",
        ),
    ]
```

src/woo_publications/publications/models.py
```diff
@@ -429,21 +431,25 @@ def revoke_own_documents(
...
-        self.bron_bewaartermijn = information_category.bron_bewaartermijn
-        self.selectiecategorie = information_category.selectiecategorie
-        self.archiefnominatie = information_category.archiefnominatie
-        self.archiefactiedatum = localdate(
-            self.registratiedatum
-            + relativedelta(years=information_category.bewaartermijn)
-        )
-        self.toelichting_bewaartermijn = information_category.toelichting_bewaartermijn
-        self.save()
+        if self.publicatiestatus != PublicationStatusOptions.concept:
```
sergei-maertens, line 434: I'd write this as a guard clause instead:

```suggestion
        if self.publicatiestatus == PublicationStatusOptions.concept:
            return
        ... # existing code
```

src/woo_publications/publications/tests/test_publications_api.py
```diff
@@ -1342,12 +1342,12 @@ def test_create_publication(self):
                 "verkorteTitel": "one",
                 "omschrijving": "Lorem ipsum dolor sit amet, "
                 "consectetur adipiscing elit.",
-                # These values will be overwritten
-                "bronBewaartermijn": "THIS VALUE WON'T BE USED",
-                "selectiecategorie": "THIS VALUE WON'T BE USED",
+                # These values will not be overwritten because the status is concept
+                "bronBewaartermijn": "THIS VALUE WILL BE USED",
+                "selectiecategorie": "THIS VALUE WILL BE USED",
                 "archiefnominatie": ArchiveNominationChoices.destroy,
```
sergei-maertens, line 1348: Do we have a similar test case to the original for creation of a published publication?

The original behaviour is still valid for published publications, so just updating the test like that creates the impression that the submitted data always overrules the computed value, but that's not the case. Please add a test for publication creation of a published publication with these fields set and assert that the automatically derived values are used.

sergei-maertens (changes requested):
I'm pleasantly surprised by the lack of complexity to realize this story :)

src/woo_publications/api/openapi.yaml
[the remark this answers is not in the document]
sergei-maertens: Press <kbd>A</kbd> to toggle annotations :)

src/woo_publications/publications/migrations/0023_alter_publication_publisher.py
sergei-maertens, line 1: note that after rebasing on main you'll have to deal with this migration - a merge migration is okay for me

src/woo_publications/publications/models.py
```diff
@@ -429,21 +431,25 @@ def revoke_own_documents(
...
+            self.archiefnominatie = information_category.archiefnominatie
+            self.archiefactiedatum = localdate(
+                self.registratiedatum
+                + relativedelta(years=information_category.bewaartermijn)
+            )
+            self.toelichting_bewaartermijn = (
+                information_category.toelichting_bewaartermijn
+            )
+
+            self.save()
```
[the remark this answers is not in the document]
bart-maykin: It is data that isn't relevant until it is published. So I think this is a nice way to reduce unnecessary database transactions. @MarcoKlerks @sergei-maertens it might even be an argument to create a database field with a timestamp from when it was published. So we can calculate the `archiefactiedatum` from that moment.

src/woo_publications/publications/tests/test_publications_api.py
bart-maykin: This is a default drf validator and this is basically saying that the field isn't allowed to be empty.
Which is different if the field isn't supplied at all which is when it will say that it is required.

Because the data is already filled out in the database this is what we need to do to test the updating of the data.

sergei-maertens: shouldn't it say `blank` instead of `null`? https://github.com/encode/django-rest-framework/blob/e454758fb6edf1dcf5aa5417a388b940c871469c/rest_framework/fields.py#L725

src/woo_publications/publications/models.py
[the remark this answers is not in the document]
MarcoKlerks: @bart-maykin Do you mean `Gepubliceerd op` from #282 and user story #283 ?

Ow, I see the latter is still in "Backlog"-status. Yikes! I'm immediately going to check the requirements, put it on "refinement" and replan it within the plateau (probably the last sprint).

src/woo_publications/publications/tests/test_publications_api.py
sergei-maertens: not when you create a published publication, right? Because then the calculation method kicks in and overwrites the api-client supplied values.

src/woo_publications/publications/tests/test_publications_api.py
bart-maykin: The dutch translation is `Dit veld mag niet leeg zijn.` so I am sure that it's fine.

src/woo_publications/publications/models.py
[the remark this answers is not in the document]
MarcoKlerks: @sergei-maertens I have changed https://github.com/GPP-Woo/GPP-publicatiebank/issues/283 in a way as you suggested.

CharString (changes requested):
2 micro adjustments :)

CharString (approved):
We'll leave that typo in the comments as an hommage to the attention spectrum :kissing_heart:

Great work.

PR #311, opened 2025-06-13
load information category management command

### LINKED ISSUE #307: The fixtures for information categories reset the archiving parameters (and omschrijving)
### Product versie / Product version

1.2.0-rc.0, main

### Omschrijf het probleem / Describe the bug

We download the available information categories from the overheid.nl waardenlijsten, but also add additional metadata that's not contained within those (such as archiving action, relations back to selectielijst, duration...) and via #305 also additional descriptions.

The information is stored in a fixture that's included in the docker image and automatically loaded on each deploy/container startup. This fixture loading results in our custom fields being overwritten with the (empty) defaults again.

Instead, we need a custom management command like we have for organisations that restore the overrides from the database after loading the fixture. It would probably be wise to work with a list/tuple of fixture-managed field names, and restore all the rest automatically so that we can future-proof this.

See https://github.com/GPP-Woo/GPP-publicatiebank/pull/305#issuecomment-2949552253

### Stappen om te reproduceren / Steps to reproduce

1. Deploy a gpp-publicatiebank instance
2. In the admin, change the archiving parameters
3. Restart the containers
4. Observe that the changes from 2. are wiped out

### Verwacht gedrag / Expected behavior

Local modifications are not reset.

bart-maykin:
Fixes #307

**Changes**

- Added fixture fields for information categories to the constants.py
- Added new management command load_information_categories
- Added wrapper to retain the original none fixture data for information category for the load_information_categories

### DIFF
--- a/src/woo_publications/metadata/tests/test_keep_none_retention_policy_field_values.py
+++ b/src/woo_publications/metadata/tests/test_keep_none_retention_policy_field_values.py
@@ -1,0 +1,41 @@
+from django.test import TestCase
+
+from woo_publications.constants import ArchiveNominationChoices
+
+from ..keep_none_retention_policy_field_values import (
+    keep_none_retention_policy_field_values,
+)
+from ..models import InformationCategory
+from .factories import InformationCategoryFactory
+
+
+class KeepOriginalNoneFixtureInformationCategoryDataTests(TestCase):
+    def test_keep_none_retention_policy_field_values(self):
+        assert not InformationCategory.objects.exists()
+        InformationCategoryFactory.create_batch(
+            10,
+            bron_bewaartermijn="bewaartermijn",
+            selectiecategorie="selectiecategorie",
+            archiefnominatie=ArchiveNominationChoices.retain,
+            bewaartermijn=5,
+            toelichting_bewaartermijn="toelichting",
+            omschrijving="omschrijving",
+        )
+
+        with keep_none_retention_policy_field_values():
+            InformationCategory.objects.update(
+                bron_bewaartermijn="",
+                selectiecategorie="",
+                archiefnominatie=ArchiveNominationChoices.destroy,
+                bewaartermijn=10,
+                toelichting_bewaartermijn="",
+                omschrijving="",
+            )
+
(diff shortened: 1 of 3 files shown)


src/woo_publications/metadata/tests/test_keep_none_retention_policy_field_values.py
```diff
@@ -0,0 +1,41 @@
...
+from ..keep_none_retention_policy_field_values import (
+    keep_none_retention_policy_field_values,
+)
+from ..models import InformationCategory
+from .factories import InformationCategoryFactory
+
+
+class KeepOriginalNoneFixtureInformationCategoryDataTests(TestCase):
+    def test_keep_none_retention_policy_field_values(self):
+        assert not InformationCategory.objects.exists()
```
CharString, line 14: nice touch

src/woo_publications/metadata/tests/test_load_information_categories.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self.assertEqual(InformationCategory.objects.count(), 10)
+
+        for obj in [ic, ic2]:
+            self.assertEqual(obj.bron_bewaartermijn, "bewaartermijn")
+            self.assertEqual(obj.selectiecategorie, "selectiecategorie")
+            self.assertEqual(obj.archiefnominatie, ArchiveNominationChoices.retain)
+            self.assertEqual(obj.bewaartermijn, 5)
+            self.assertEqual(obj.toelichting_bewaartermijn, "toelichting")
+            self.assertEqual(obj.omschrijving, "omschrijving")
```
CharString, line 90: Isn't it clearer to setup this test as
1. call command
2. make changes
3. call command again
4. assert changes have been kept?

Or am I misunderstanding the testcase?

bart-maykin: You are misunderstanding the testcase. I want to test that if some existing data would not overwrite the user input data. So the test is:

1. Create data which correspond in the fixture.
2. call the command.
3. assert that the user data didn't change.

CharString (approved):

PR #332, opened 2025-07-03
Filter out revoked publications in document admin page

### LINKED ISSUE #330: Filter out revoked publicaties in document admin
Filter out revoked publicaties in the dropdown in the document admin panel.
This way users can't accidently try to create a new document for a revoked publication which triggers an error

bart-maykin:
Fixes #330

**Changes**

Alter the document admin's publicatie queryset to filter out revoked publications

### DIFF
--- a/src/woo_publications/publications/admin.py
+++ b/src/woo_publications/publications/admin.py
@@ -684,6 +684,13 @@ class DocumentAdmin(AdminAuditLogMixin, admin.ModelAdmin):
                 )
             )
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        if db_field.name == "publicatie":
+            kwargs["queryset"] = Publication.objects.filter(
+                ~models.Q(publicatiestatus=PublicationStatusOptions.revoked)
+            )
+        return super().formfield_for_foreignkey(db_field, request, **kwargs)
+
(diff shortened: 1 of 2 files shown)


src/woo_publications/publications/admin.py
```diff
@@ -698,6 +698,13 @@ def delete_queryset(
                 )
             )
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        if db_field.name == "publicatie":
+            kwargs["queryset"] = Publication.objects.filter(
+                ~models.Q(publicatiestatus=PublicationStatusOptions.revoked)
+            )
```
sergei-maertens, line 705: I think this drops potential application of `limit_choices_to` at the model field level, and may drop the ordering from the admin. Probably safer to repeat the default implementation and modify that:

```suggestion
            default_qs = self.get_field_queryset(db, db_field, request)
            assert default_qs is not None
            kwargs["queryset"] = default_qs.exclude(publicatiestatus=PublicationStatusOptions.revoked)
```

Also note the usage of `.exclude` instead of manually building that using `~models.Q(...)` which is much less readable.

sergei-maertens (changes requested):

sergei-maertens (approved):
I think I would've just `assert default_queryset is not None` instead of the fallback, but the explanation clarifies things sufficiently.

PR #358, opened 2025-07-15
:arrow_up: Upgrade all backend dependencies

sergei-maertens:
... to their latest version.

The result of ./bin/compile_dependencies.sh --upgrade.


requirements/base.txt
```diff
@@ -209,7 +208,7 @@ django-upgrade-check==1.1.0
     # via
     #   -r requirements/base.in
     #   open-api-framework
-djangorestframework==3.15.2
+djangorestframework==3.16.0
```
sergei-maertens, line 211: -> check changelog for breaking changes

PR #408, opened 2026-01-16
Strip metadata

### LINKED ISSUE #386: Strip embedded metadata van PDF-bestanden
_Zie bronnen vermeld op de epic https://github.com/GPP-Woo/.github/issues/79_

**Acceptatie criteria**
- Als een PDF-bestand ontvangen of opgehaald wordt (zie #274), verwijder dan **alle** embedded metadata. 
~de volgende embedded metadata:~
  - ~`/Author`~
  - ~`/Title`~
  - ~`/Subject`~
  - ~`/Keywords`~
  - ~`<dc:creator>`~
  - ~`<pdf:Keywords>`~

**Acties**

- [ ]

bart-maykin:
Fixes #386

**Changes**

- Added document model field to indicate when a document's metadata got stripped.
- Added mimetype detection for file upload.
- Created celery task to strip and re-upload the document in the documents api (this requires a delete step because updating an existing document in the documents api doesn't create "bestandsdelen")
- Ensures that pdf files which doesn't have their metadata stripped aren't indexed in the search component and can't be downloaded
- Automatically call the meta data strip task during document indexing if it's a pdf file which isn't stripped yet.
- Moved certain bits of code to be able to re-use in the strip metadata task.


src/woo_publications/publications/tasks.py
```diff
@@ -18,10 +25,84 @@
...
+                writer = PdfWriter(temp_file)
+
+                # strip all metadata
+                writer.metadata = None
+
+                # save file
+                _, stripped_file = writer.write(stripped_file)
+
+            # delete document in documents api
+            client.destroy_document(uuid=document.document_uuid)
```
CharString: nvm... client wordt nog gebruikt.

CharString (changes requested):
Nice. Nog wat tests voor de nieuwe functionaliteit.

sergei-maertens (changes requested):

sergei-maertens (changes requested):

src/woo_publications/publications/tasks.py
```diff
@@ -22,6 +27,105 @@
...
+        # set stripped file back to the starting bytes
+        stripped_file.seek(0)
+
+        is_completed = False
+        # upload file parts
+        for part in file_parts:
+            file_part = stripped_file.read(part.size)
+            is_completed = document.upload_part_data(uuid=part.uuid, file=file_part)
+
+        return is_completed
```
sergei-maertens: this will always result in `return True` right, there's no reason you'd stop before the last file part.

Is there a point/reason for the task to return a result then?

src/woo_publications/publications/tests/test_index_document_task.py
```diff
@@ -67,13 +67,42 @@ def test_index_skipped_for_incomplete_uploads(self):
...
+        )
+        bestandsnaam = DocumentFactory.create(
+            publicatiestatus=PublicationStatusOptions.published,
+            upload_complete=True,
+            bestandsnaam="test.pdf",
+        )
+
+        with (
+            self.subTest("check from bestandsformaat"),
+            self.captureOnCommitCallbacks(execute=True),
```
sergei-maertens: why do the on commit callbacks need to be captured/executed?

src/woo_publications/publications/tests/test_strip_all_files_command.py
```diff
@@ -0,0 +1,115 @@
...
+            stdout=out,
+            no_color=True,
+        )
+
+        self.assertEqual(out.getvalue(), "Search API services not configured.\n")
+        mock_strip_pdf.assert_not_called()
+
+    @requests_mock.Mocker()
+    def test_incorrect_url(self, mock_strip_pdf: MagicMock, m):
+        m.register_uri(requests_mock.GET, requests_mock.ANY, status_code=404)
```
sergei-maertens: I don't see why we need to `requests_mock` this instead of using plain VCR - you can use the `with self.vcr_raises(...)` approach here too, and/or a base URL like `https://example.com/<some-uuid>` that's pretty certain to 404.

The first suggestion will highlight a case you missed in the command probably, where DNS/IO errors are not caught currently.

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #450, opened 2026-06-11
Fixes #449 -- make hardcoded uwsgi flags overridable via environment variables

### LINKED ISSUE #449: uwsgi flags hardcoded in docker_start.sh ignore UWSGI_* environment variables
`bin/docker_start.sh` passes `--http-timeout`, `--post-buffering`, `--buffer-size` and `--harakiri` as hardcoded CLI flags. uwsgi gives CLI flags precedence over its `UWSGI_*` environment variables, so e.g. `UWSGI_HARAKIRI=1860` set on the container is silently ignored — unlike `UWSGI_PORT`, `UWSGI_PROCESSES` and `UWSGI_THREADS`, which the script already reads from the environment.

Deployments need to tune these without forking the image, e.g. to keep `harakiri` above the reverse proxy's read timeout (a clean 504 instead of a worker SIGKILL), or to adjust `post-buffering` as a workaround for unbit/uwsgi#2706 (spurious worker kills during large uploads when `threads > 1`).

Suggestion: read the four options from `UWSGI_HTTP_TIMEOUT`, `UWSGI_HARAKIRI`, `UWSGI_POST_BUFFERING` and `UWSGI_BUFFER_SIZE`, defaulting to the current values, following the script's existing pattern. No behaviour change for anyone not setting them. PR follows.

RamsesRodenburg:
Fixes #449

**Changes**

`bin/docker_start.sh`: export `UWSGI_PROCESSES`, `UWSGI_THREADS`, `UWSGI_HTTP_TIMEOUT`, `UWSGI_HARAKIRI`, `UWSGI_POST_BUFFERING` and `UWSGI_BUFFER_SIZE` with the previous values as defaults, and drop the matching explicit CLI flags — uwsgi picks the env vars up natively, so the infra-layer can override any of them without code changes. Effective defaults are unchanged.

### DIFF
--- a/bin/docker_start.sh
+++ b/bin/docker_start.sh
@@ -12,6 +12,10 @@ fixtures_dir=${FIXTURES_DIR:-/app/fixtures}
 uwsgi_port=${UWSGI_PORT:-8000}
 uwsgi_processes=${UWSGI_PROCESSES:-4}
 uwsgi_threads=${UWSGI_THREADS:-1}
+uwsgi_http_timeout=${UWSGI_HTTP_TIMEOUT:-1800}
+uwsgi_harakiri=${UWSGI_HARAKIRI:-1800}
+uwsgi_post_buffering=${UWSGI_POST_BUFFERING:-8192}
+uwsgi_buffer_size=${UWSGI_BUFFER_SIZE:-65535}
 
 mountpoint=${SUBPATH:-/}
 
@@ -68,7 +72,7 @@ exec uwsgi \
     --master \
     --http :$uwsgi_port \
     --http-keepalive \
-    --http-timeout=1800 \
+    --http-timeout=$uwsgi_http_timeout \
     --manage-script-name \
     --mount $mountpoint=woo_publications.wsgi:application \
     --static-map /static=/app/static \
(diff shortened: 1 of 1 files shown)


sergei-maertens (changes requested):
uwsgi already reads envvars with the `UWSGI_*` prefix by default. I'd take a different approach here and make sure that default envvars are specified, and remove the matching explicit flags (because those override envvars). The pattern we use in other projects is for example:

```bash
export UWSGI_PROCESSES=${UWSGI_PROCESSES:-4}
```
Otherwise this creates the impression too much that uwsgi doesn't support envvars by default, and I do prefer being able to just set envvars at the infra-layer without needing code changes for every other modification.

You can take a look at the Open Forms script for some better practices: https://github.com/open-formulieren/open-forms/blob/main/bin/docker_start.sh

@bart-maykin FYI

RamsesRodenburg:
Done in 42d7e7b — defaults are now exported as `UWSGI_*` envvars and the matching explicit flags are removed.

sergei-maertens (approved):
Nice, thanks!

Some more small cosmetic feedback, can you:

* include the issue reference in your second commit, or
* squash the second commit into the first

This way we can still trace back where this changeset came from in a year :-)

RamsesRodenburg:
Thanks! Would you be okay if I perform Squash-and-merge of the PR? That should result in a single commit mentioning fix for issue#449...?

sergei-maertens:
> Thanks! Would you be okay if I perform Squash-and-merge of the PR? That should result in a single commit mentioning fix for issue#449...?

Yep, in this case it's okay, for regular PRs where a bugfix/feature is implemented with multiple commits (regression test + fix as separate commits for example), we favour a plain merge so that we can easily still see the reasoning for the individual commits and can check out individual commits to validate things :-)

PR #454, opened 2026-06-11
Fixes #453 -- recycle celery pool children by default to bound worker memory

### LINKED ISSUE #453: celery worker pool child is never recycled, accumulating memory until the pod is OOMKilled
**Product versie / Product version**: 2.1.0 (celery 5.6.2)

After publishing two batches of ~100 small documents (~1 GB combined), the worker pod was OOMKilled at its 1200MiB limit — the single prefork pool child had grown to ~1.1 GiB:

```
Memory cgroup out of memory: Killed process 1470272 (celery) total-vm:1507368kB, anon-rss:1143608kB, file-rss:15788kB
```

`bin/celery_worker.sh` starts the worker without `--max-tasks-per-child` / `--max-memory-per-child`, so with the default concurrency of 1 a single long-lived child executes every task. CPython rarely returns freed memory to the OS, so RSS accumulates across document-processing tasks until whatever memory limit applies is hit — the worker-side analogue of what `max-requests` mitigates for uwsgi on the web tier.

Because the CLI is invoked via `celery` (Click, `auto_envvar_prefix="CELERY"`), the `CELERY_WORKER_MAX_TASKS_PER_CHILD` / `CELERY_WORKER_MAX_MEMORY_PER_CHILD` environment variables already bind to the corresponding worker options (verified on 5.6.2), so deployments can opt in today. Proposal, in the same spirit as #449: export a recycling default in `celery_worker.sh` so installations are protected out of the box:

```bash
export CELERY_WORKER_MAX_TASKS_PER_CHILD=${CELERY_WORKER_MAX_TASKS_PER_CHILD:-50}
```

Pool children are recycled between tasks, so no work is lost. PR ready.

RamsesRodenburg:
Closes #453

Exports an overridable `CELERY_WORKER_MAX_TASKS_PER_CHILD=50` default in `bin/celery_worker.sh`, bounding the RSS growth of the long-lived pool child described in #453. Celery binds `CELERY_WORKER_*` environment variables to worker CLI options natively (`auto_envvar_prefix="CELERY"`), so no explicit flag is needed and deployments can override or disable (`0`) it — same pattern as #450.
Imposing a default, non-zero maxRequests for the Celery worker prevents workers from eventually exceeding their memory limit - getting the in-flight request OOMKilled and the upload failed. Or end up consuming all memory. ;)

### DIFF
--- a/bin/celery_worker.sh
+++ b/bin/celery_worker.sh
@@ -5,6 +5,14 @@ set -e
 LOGLEVEL=${CELERY_LOGLEVEL:-INFO}
 CONCURRENCY=${CELERY_WORKER_CONCURRENCY:-1}
 
+# Celery reads CELERY_WORKER_* environment variables natively (Click
+# auto_envvar_prefix="CELERY"), so any worker option can be overridden from the
+# environment. Default to recycling pool children periodically so memory
+# released by completed tasks is returned to the OS instead of accumulating in
+# a long-lived child process. Recycling happens between tasks; no work is lost.
+# CELERY_WORKER_MAX_MEMORY_PER_CHILD (in KiB) can be set as a companion limit.
+export CELERY_WORKER_MAX_TASKS_PER_CHILD=${CELERY_WORKER_MAX_TASKS_PER_CHILD:-50}
+
 QUEUE=${CELERY_WORKER_QUEUE:=celery}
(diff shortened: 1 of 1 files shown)


bart-maykin (approved):

bart-maykin:
Remade the pr https://github.com/GPP-Woo/GPP-publicatiebank/pull/469 because of lack of authority on forked repo.

